PR #327, opened 2021-02-23
#219 connection investments ptdf

DillonJ:
Hi folks,

So this PR implements PTDF friendly investments. The issue is documented in #216 but I will try to summarise here.

The formulation changes as follows:

DC load flow can be implemented using power transfer distribution factors (PTDFs). The flow on connection l using PTDFs is expressed as :

image
where PTDF(l,n) is the fraction of an injection at n that flows on l.

This is implemented in the constraint connection_flow_ptdf

Line outage distribution factors (LODF) express the fraction of the pre-outage flow on line(connection) k that flows on line l if line K is removed. In the context of investments, we can use LODFs to calculate the impact on network flows if investment candidate connection c is present or not.

To do this, we re-interpret the results of the above equation as the flows in the intact network, with all candidate connections invested in. We rename the constraint to `connection_intact_flow_ptdf` and we create a new variable that represents the intact network flows `connection_intact_flow`.

Then the actual flow on connection l is the `connection_intact_flow` plus the additional flow for every candidate connection not invested in. This is expressed in constraint connection_flow_intact_flow as follows:

image. 

The term intact_flow(c) - flow(c) evaluates to intact_flow(c) if the connection is not invested in - otherwise the subsequent constraints ensure that intact_flow(c) - flow(c) is zero if connection c is invested in.

The following two constraints ensure that if connection c is invested in the flow on connection c is equal to intact_flow, otherwise it is zero. This is implemented in constraints candidate_connection_flow_intact_flow, connection_flow_capacity and candidate_connection_flow_intact_flow2 as follows:

image

This formulation avoids a multiplication of the investment variable and a flow variable.


DillonJ:
@manuelma perhaps you could take a look? I know you're busy though!

manuelma:
Will do! This one's exciting...

mihlema:
I think @iasonask might also have some background here.

manuelma:
> The term intact_flow(c) - flow(c) evaluates to intact_flow(c) if the connection is invested in - otherwise the subsequent constraints ensure that intact_flow(c) - flow(c) is zero if connection c is invested in.

Just to be sure, this should be ...The term intact_flow(c) - flow(c) evaluates to intact_flow(c) if the connection is **not** invested in...

manuelma:
flow(c) = {intact_flow(c) if c is invested in, zero otherwise} ?

DillonJ:
> > The term intact_flow(c) - flow(c) evaluates to intact_flow(c) if the connection is invested in - otherwise the subsequent constraints ensure that intact_flow(c) - flow(c) is zero if connection c is invested in.
> 
> Just to be sure, this should be ...The term intact_flow(c) - flow(c) evaluates to intact_flow(c) if the connection is **not** invested in...

You're right - I corrected the original comment

manuelma:
I would suggest

flow(c) > intact_flow(c) + (1 - connection_invested_available(c)) * capacity(c)

To avoid that 100000.

DillonJ:
> I would suggest
> 
> flow(c) > intact_flow(c) + (1 - connection_invested_available(c)) * capacity(c)
> 
> To avoid that 100000.

I considered that - but are we guaranteed there is a capacity defined for that flow?

manuelma:
> I considered that - but are we guaranteed there is a capacity defined for that flow?

With the 100.000, it results in intact_flow(c) < 100.000 when c not invested in... Isn't that dangerous? Maybe take the julia max float then?

src/constraints/constraint_connection_flow_capacity_strict.jl
```diff
@@ -0,0 +1,125 @@
...
+Limit the maximum in/out `connection_flow` of a `connection` for all `connection_flow_capacity_strict` indices.
+
+Check if `connection_conv_cap_to_flow` is defined. The `connection_capacity` parameter is used to constrain the
+"average power" (e.g. MWh/h) instead of "instantaneous power" (e.g. MW) of the `connection`.
+For most applications, there isn't any difference between the two. However, for situations where the same `connection`
+handles `connection_flows` to multiple `nodes` with different temporal resolutions, the constraint is only generated
+for the lowest resolution, and only the average of the higher resolution `connection_flow` is constrained.
+If instantaneous power needs to be constrained as well, defining the `connection_capacity` separately for each
+`connection_flow` can be used to achieve this.
+"""
```
manuelma, line 32: Are new functions docstrings up to date? Not just copy/pasted? (Just making sure...)

src/constraints/constraint_connection_flow_intact_flow.jl
```diff
@@ -0,0 +1,131 @@
...
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+# GNU Lesser General Public License for more details.
+#
+# You should have received a copy of the GNU Lesser General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#############################################################################
+
+"""
+    add_constraint_connection_flow_capacity!(m::Model)
```
manuelma, line 21: This doesn't match the function name.

src/constraints/constraint_connection_flow_intact_flow.jl
```diff
@@ -0,0 +1,131 @@
...
+                    t=t_in_t(m; t_long=t),
+                );
+                init=0,
+            )*duration(t) 
+        ) for (conn, ng, s, t) in constraint_connection_flow_intact_flow_indices(m)
+    )
+end
+
+"""
+    constraint_connection_flow_capacity_indices(m::Model; filtering_options...)
```
manuelma, line 81: Doesn't match the function name.

src/SpineOpt.jl
```diff
@@ -128,7 +129,13 @@ include("constraints/constraint_node_state_capacity.jl")
 include("constraints/constraint_ratio_unit_flow.jl")
 include("constraints/constraint_ratio_out_in_connection_flow.jl")
 include("constraints/constraint_connection_flow_capacity.jl")
-include("constraints/constraint_connection_flow_ptdf.jl")
+include("constraints/constraint_connection_flow_capacity_strict.jl")
+include("constraints/constraint_connection_intact_flow_capacity.jl")
+include("constraints/constraint_connection_flow_intact_flow.jl")
```
manuelma, line 134: Maybe we can rename this to `constraint_connection_intact_flow` to avoid repetition of the word flow?

src/SpineOpt.jl
```diff
@@ -128,7 +129,13 @@ include("constraints/constraint_node_state_capacity.jl")
...
+include("constraints/constraint_connection_intact_flow_ptdf.jl")
+include("constraints/constraint_connection_intact_flow_ptdf_in_out.jl")
+include("constraints/constraint_candidate_connection_flow_intact_flow.jl")
+include("constraints/constraint_candidate_connection_flow_intact_flow2.jl")
```
manuelma, line 138: Would you consider:

constraint_candidate_connection_flow_intact_flow -> constraint_candidate_connection_flow_lower_bound

constraint_candidate_connection_flow_intact_flow2 -> constraint_candidate_connection_flow_upper_bound

test/constraints/constraint_connection.jl
```diff
@@ -820,5 +812,5 @@
             @test _is_constraint_equal(observed_con, expected_con)     
             return       
         end
-    end
+    end    
 end
```
manuelma, line 816: Thanks for fixing the tests. Would you consider writing tests for the new constraints? Maybe after we merge?

manuelma (changes requested):
I feel like reviewing a paper with this new system. I'm requesting a minor revision. In general, I like the approach and look forward to use in my case study.

DillonJ:
> > I considered that - but are we guaranteed there is a capacity defined for that flow?
> 
> With the 100.000, it results in intact_flow(c) < 100.000 when c not invested in... Isn't that dangerous? Maybe take the julia max float then?
> 
> **Edit**: `typemax(Int)` works inside `JuMP.@constraint`.

Actually, I updated pre-processing recently to copy the capacity "to the other side" when creating the additional relationships - so we are more than likely to have a capacity... we could just use the max int (or whatever) if there is no capacity defined I suppose

test/constraints/constraint_connection.jl
DillonJ: Of course, that is the plan... I'm working on them

src/SpineOpt.jl
DillonJ: that's nicer all right

manuelma:
> I would suggest

> flow(c) > intact_flow(c) + (1 - connection_invested_available(c)) * capacity(c)

> To avoid that 100000.

> I considered that - but are we guaranteed there is a capacity defined for that flow?

So I suggest to use `typemax(Int)` instead.

If the flow capacity is not defined, what happens to the constraint just above? flow(c) <= connections_invested_available(c) * cap(c)?

Also, `connections_invested_available` can only be zero or one? If yes, why the plural?

src/SpineOpt.jl
DillonJ: I was trying to capture that the constraint is capturing the relationship between the two flows - perhaps `constraint_connection_flow_to_intact_flow` does that read better?

src/constraints/constraint_connection_flow_capacity_strict.jl
DillonJ: oops... they may have been copy/pasted lol - will fix them

DillonJ:
So @manuelma I believe I have addressed most of the issues. Now I use connection_capacity instead of 100000. If there is no capacity, the constraint falls back on 100000 but I made changes to pre-processing for bi-directional lossless lines to ensure that this should never be needed. I can't use typemax(Int) but it gets multiplied by the timeslice duration and we'd have an overflow otherwise.

src/data_structure/preprocess_data_structure.jl
```diff
@@ -177,6 +166,54 @@ function generate_direction()
...
+"""
+
+function process_loss_bidirectional_capacities()
+    for c in connection(connection_type=:connection_type_lossless_bidirectional)
+        conn_capacity_param = nothing
+        found_from = false
+        for (n, d) in connection__from_node(connection=c)
+            if haskey(connection__from_node.parameter_values[(c, n, d)], :connection_capacity)
+                conn_capacity_param = connection__from_node.parameter_values[(c, n, d)][:connection_capacity]
+                found_from = true
```
manuelma, line 184: `haskey()` before accessing the dictionary means two dict lookups. Would you consider using `get` instead? (it only requires one lookup)

```julia
found_value = get(dict, key, nothing)
if found_value != nothing
    # use found_value directly
else
    # not in dictionary
end

```

DillonJ:
Final test implemented - I believe there is now a test for everything I've added

src/data_structure/preprocess_data_structure.jl
DillonJ: ah, ok - good to know

manuelma: I see that we look for connection_capacity only in parameter values, not in default values. Is that important?

We have the `indices` function that can collect keys where the value is different than `nothing`. It also accepts filters: 
```julia

for (c, n, d) in indices(connection_capacity; node=n, direction=d)
   ...
end
```

Maybe we can use it?

Also, I'm not a big fan of accessing and modifying the fields of `Parameter` values directly. SpineInterface should provide a function that does that. What do you think of something like `set_value!(parameter, class, index, value)`?

manuelma: Did you add any tests for this code? If yes, I think we can merge and I can do a bit of refactoring later...

DillonJ: > I see that we look for connection_capacity only in parameter values, not in default values. Is that important?
> 
> We have the `indices` function that can collect keys where the value is different than `nothing`. It also accepts filters:
> 
> ```julia
> for (c, n, d) in indices(connection_capacity; node=n, direction=d)
>    ...
> end
> ```
> 
> Maybe we can use it?
> 
> Also, I'm not a big fan of accessing and modifying the fields of `Parameter` values directly. SpineInterface should provide a function that does that. What do you think of something like `set_value!(parameter, class, index, value)`?

There are two issues here... if I use indices(connection_capacity, ...) I don't know if what I'm getting back is on `connection__from_node` or `connection__to_node` - the way I have written it is the only way I could figure out to know on which relationships and which indices exactly I need to set connection_capacity.

The other issue is that I have no idea what the parameter may look like - when I go to the dicts directly I can just copy the parameter_value and it doesn't matter if it's a static value or a timeseries or whatever... so if we were to add support in SpineInterface for this sort of thing - we would also need `get_value_raw(parameter, class, index, value)` (or whatever)?

DillonJ: > Did you add any tests for this code? If yes, I think we can merge and I can do this bit of refactoring later...

You mean the pre-processing code? Not yet - but I have implemented tests for all the new constraints. I did test that the new pre-processing code works correctly, and I will write some specific tests for it. I don't have time today - but perhaps we could merge anyway and I will do it on Monday? I always get a little anxious of slipping behind and having to deal with merge conflicts so it would be nice to get it in if we could?

manuelma: Let's merge and 'worry' about this later. You made fair points, and it looks like things need to get thought trough in SpineInterface. There's no point in delaying this PR because of that.

Everything else looks great, so I will finish my review so you can merge?

manuelma (approved):

PR #330, opened 2021-02-24
#191 fixing bug

mihlema:
availability factor had to be replaced with unit_conv_cap_to_flow


Tasqu (commented):
I agree, it seems the `unit_availability_factor` should be accounted for by `constraint_units_available` already, so currently this seems to be accounting for it twice.

Issue #319, opened 2021-02-10
Current implementation of representative days including seasonal storages and next steps


mihlema:
This issue serves as a place to discuss the current implementation of rep. days with seasonal storages, and next steps and possible improvements.\
### The following elements are required database side:
- `model` object
    - parameter: `use_representative_days`:	`use_representative_days_w_seasonal_storage`
- `temporal_block` object
    - (at least) 1 `temporal_block` covering the non-representative period (e.g. `1year`)
        - parameter: `ordering_periods`: Map 
    e.g. `{"type": "map", "index_type": "date_time", "data": 
[["2017-01-01T00:00:00", {"type": "date_time", "data": "2017-01-02T00:00:00"}], 
["2017-01-01T01:00:00", {"type": "date_time", "data": "2017-01-02T01:00:00"}], 
...
["2017-12-31T23:00:00", {"type": "date_time", "data": "2017-02-26T23:00:00"}]]}`
        - the `ordering_periods` mapping parameter maps each timestep of the non-representative days to the corresponding timesteps of the representative days<sup>1</sup>
    - (at least) 1 `temporal_block` representing the representative period (e.g. `Rep-day1`)
        - Parameter: Ordering_periods
---
<sup>1</sup> I'm currently revising this to a mapping parameter, were only the start of a non-representative period (e.g. the start of each day) is required, mapping to a representative temporal_block.

### Code-side this translates to the following adjustments:
- `Variable_indices` functions:
    - indices function do now have a temporal_block keyword. The default for this temporal_block argument is `setdiff(temporal_block(), indices(ordering_periods))`, meaning that the default for the indices functions are the representative indices 
(Note: this doesn't have an impact for optimizations without representative days, as the `ordering_periods` parameter doesn't exist in this case, and the temporal_block will be equal to `anything`)
    - The only exceptions are variables, that we explicitly want to explicitly model over the whole timeframe, in particular the `node_state` variable. A node can be made a seasonal storage by connecting it to both, the rep.day temporal_blocks *and* the non-representative period (e.g. the temporal_block `1year`) <sup>2</sup>.
    - Additionally, the `dynamic_unit/connection...indices` functions receive an extra temporal_block argument, with the default being `anything` -> meaning that all timeslices (representative *and* non-representative) are being called (the problematic with this is explained further below)
- `add_variable!` function:
    - the generation of variables (and variable references) remains the same (keeping in mind that this function calls the default `variable_indices` function, which defaults to the representative indices)
[...]

PR #335, opened 2021-02-26
WIP: #319 Seasonal storage/rep. days

mihlema:
Current implementation allows us to have representative days + seasonal storages. node_state variables, investment variables are considered to be always non-representative, i.e. always exist for the full optimization period, while all others are representative, i.e. mapped for non-representative periods to their corresponding representative period.
Mapping of the variables happens in `variable_common`. The filtering of `rep` and `non-rep` temporal_blocks happens in the variable_indices functions. 
Additional features:
- dynamic constraint: dynamic constraints are only generated for rep. days, but for their overlap to previous period, we also take into account the representative period of the previous period. See e.g. `unit_dynamic_indices`
- to ease the implementation of multiple temporal_blocks, a temporal_block group can be used to connect to node. Members function is called within the code
Known issues:
- as stated before, node_state balance is always assumed to be non-representative,i.e. created for the full year. At the moment it's not possible to combine non-seasonal and seasonal storages. Would we need this?
- the node_injection constraint is created for the enitre year. However, for e.g. nodes without storages, it could also simply be created for only rep. days. I haven't found a way to implement this is a nice way
- in the `node_injection` constraint, the representative demand of the current timeslice needs to be called. The current implementation is working, but not very pretty.
- currently the temporal_block weights are summed up. This is a bug and needs to be changed to prod(...) . Currently this is not supported by Spineopt and interface


src/run_spineopt.jl
```diff
@@ -324,6 +324,31 @@ function optimize_model!(m::Model; log_level=3, calculate_duals=false, mip_solve
     # NOTE: The above results in a lot of Warning: Variable connection_flow[...] is mentioned in BOUNDS,
     # but is not mentioned in the COLUMNS section. We are ignoring it.
     @timelog log_level 0 "Optimizing model $(m.ext[:instance])..." optimize!(m)
+    if termination_status(m) == MOI.INFEASIBLE
+        compute_conflict!(m)
```
manuelma, line 328: This is nice. Would you consider putting this code in its own function, and call it from here?

Also, just to be sure this is not related to the current MR? If not, would you consider committing it to master directly?

src/variables/variable_common.jl
```diff
@@ -52,9 +52,33 @@ function add_variable!(
...
+"""
+    map_variable!(v::Dict{VariableRef}, indices_all::Function)
+
+Extends variable Dictionary such that indexing by a non representative
+time slice returns the variable with the corresponding representative
+time slice.
+"""
+function map_variable!(m::Model, var::Dict, var_indices::Function)
+    # TODO: should probably rename "ordering_periods" to
+    # "time_slice_mapping"
```
manuelma, line 70: I think `representative_periods` is a nice name. "Ordering" doesn't sound correct to me, sounds like something else.

`representative_periods` or `representative_time_slices` sounds good. `time_slice_mapping` is too general and doesn't say anything about 'representative'.

src/variables/variable_unit_flow.jl
```diff
@@ -35,18 +35,21 @@ function unit_flow_indices(
...
         (unit=u, node=n, direction=d, stochastic_scenario=s, t=t)
         for
-        (u, n, d, tb) in
-        unit__node__direction__temporal_block(unit=unit, node=node, direction=direction, _compact=false)
+        (u, n, d, temporal_block) in
+        unit__node__direction__temporal_block(unit=unit, node=node, direction=direction,temporal_block=temporal_block,_compact=false)
         for
         (n, s, t) in
-        node_stochastic_time_indices(m; node=n, stochastic_scenario=stochastic_scenario, temporal_block=tb, t=t)
+        node_stochastic_time_indices(m; node=n, stochastic_scenario=stochastic_scenario, temporal_block=temporal_block, t=t)
```
manuelma, line 50: Why this change? I think it's more consistent to name the iteration variable `tb`, as it was before...

src/variables/variable_common.jl
```diff
@@ -52,9 +52,33 @@ function add_variable!(
...
+    for ind in setdiff(var_indices(m;temporal_block=anything),var_indices(m))
+        # Get indices which aren't time slices
+        Keys = [k for k in keys(ind) if !(typeof(ind[k]) <: TimeSlice)]
+        Values = [ind[k] for k in Keys]
+        non_t_slice_ind = (; zip(Keys, Values)...) #this gets everything, but the timeslice index...
+        ind_rep = first(var_indices(m;non_t_slice_ind...,t=rep_time_slices(m)[to_time_slice(m,t=ind.t)]))
```
manuelma, line 76: Why do we need to call `to_time_slice` here? isn't always the case that `ind.t` is already in the model? I don't see how else it would have made it to a variable index (maybe I'm missing something?)

Because if I'm not mistaken `to_time_slice(m,t=ind.t) == t.ind.t`

src/objective/variable_om_costs.jl
```diff
@@ -30,6 +30,7 @@ function variable_om_costs(m::Model, t1)
         expr_sum(
             +unit_flow[u, n, d, s, t] *
             duration(t) *
+            sum(weight(temporal_block=blk) for blk in t.blocks) *
```
manuelma, line 33: I have a small suggestion: replace `t.blocks` by `blocks(t)`. It's more api-friendly than accessing fields directly.

src/variables/variable_common.jl
```diff
@@ -52,9 +52,33 @@ function add_variable!(
             ind => _variable(m, name, ind, lb, ub, bin, int)
             for ind in indices(m; t=vcat(history_time_slice(m), time_slice(m)))
         )
+    if use_representative_days(model=m.ext[:instance]) == :use_representative_days_w_seasonal_storage
```
manuelma, line 55: Is the spine_opt_template.json updated at this point? I think we'd need to add the use_representative_days parameter together with the possible 'enumerated' values. Which others would we have apart from use_representative_days_w_seasonal_storage? is the `w_seasonal_storage` part really necessary? Maybe it could be just a true/false thing.

src/data_structure/temporal_structure.jl
```diff
@@ -344,6 +345,28 @@ function to_time_slice(m::Model; t::TimeSlice)
...
+            rep_blk = ordering_periods(temporal_block=blk, inds=t_start_real)
+            t_start_real_i = t_start_real
+            for t in time_slice(m, temporal_block=temporal_block(rep_blk))
+                rep_dict[to_time_slice(m,t=TimeSlice(t_start_real_i,t_start_real_i + _model_duration_unit(m.ext[:instance])(duration(t))))] = t
+                t_start_real_i = t_start_real_i + _model_duration_unit(m.ext[:instance])(duration(t))
+            end
+        end
+    end
+    m.ext[:temporal_structure][:rep_day_mapping] = rep_dict
+    m.ext[:temporal_structure][:non_rep_ind] = keys(rep_dict)
```
manuelma, line 365: I don't see `:non_rep_ind ` being used anywhere else?

src/variables/variable_common.jl
```diff
@@ -52,9 +52,33 @@ function add_variable!(
...
+        map_variable!(m, m.ext[:variables][name], indices)
+    end
     ((bin != nothing) || (int != nothing)) && push!(m.ext[:integer_variables], name)
 end
 
+"""
+    map_variable!(v::Dict{VariableRef}, indices_all::Function)
```
manuelma, line 62: maybe the name could be `replace_non_representative_periods!(...)` or `map_periods!(...)`? I don't know, `map_variable!` sounds a bit odd. Maybe `remap_variable!` ???

src/data_structure/temporal_structure.jl
```diff
@@ -283,6 +283,7 @@ function generate_temporal_structure!(m::Model)
     _generate_current_window!(m::Model)
     _generate_time_slice!(m::Model)
     _generate_time_slice_relationships!(m::Model)
+    rep_time_slice_mapping(m::Model)
```
manuelma, line 286: I think, for consistency this should be called `_generate_representative_time_slice_mapping()`.

Also, I'm not sure I'm a fan of abbreviating `representative` to `rep` in general?

src/data_structure/temporal_structure.jl
```diff
@@ -344,6 +345,28 @@ function to_time_slice(m::Model; t::TimeSlice)
...
 end
 
+"""
+    rep_time_slice_mapping(m::Model)
+Generate an `Array` mapping all non-representative to representative time-slices
+"""
+function rep_time_slice_mapping(m::Model)
+    rep_dict=Dict()
+    for blk in indices(ordering_periods)
+        for t_start_real in ordering_periods(temporal_block=blk).indexes
```
manuelma, line 355: We'd need a check that `ordering_periods` is a TimeSeries to be able to use `indexes`. But also I'm not a fan of accessing fields directly...

manuelma:
I went through most of the code and I really like it. I suggested a bunch of renaming ideas and also asked a couple of minor questions that would be good to answer before merging, but apart from that, since test pass (as I believe they do) it's ok to merge.

We probably want to find time to add tests for the new functionality so we don't accidentally break in future pushes.

src/run_spineopt.jl
mihlema: Yes, sorry this one was supposed to come in a separate MR. I'll try to remove it from this one and merge it separately. But it's indeed really cool, already saved me a bunch of time.

src/objective/variable_om_costs.jl
mihlema: Yes that looks better. Also, related to the original MR message:
> currently the temporal_block weights are summed up. This is a bug and needs to be changed to prod(...) . Currently this is not supported by Spineopt and interface
I have tried to use prod(..) or Iterators.prod(..) here, but the iteration method seems to fail here. Do you have an idea on that? I guess alternatively reduce(*,...) could work

src/variables/variable_common.jl
mihlema: Yes, I think we could have a boolean parameter `use_representative_days_w_seasonal_storage`. I think the `w_seasonal_storage` doesn't look really nice, but I think it'd be important to indicate it here. (For "normal" representative days we don't need the whole mapping thing).
Alternatively, we can also trigger the mapping by the existence of the `ordering_periods` parameter, I think. Then we wouldn't need the `use_representative_days_w_seasonal_storage` parameter at all.

mihlema:
Thanks manuel for reviewing. I have addressed your suggestions and removed the "find conflicts functionality" and will submit another merge request on it. This should be good now.

PR #342, opened 2021-03-17
WIP: Additional run_spineopt functionality

mihlema:
- for infeasible, direct models, run_spineopt now return the conflicting constraints
- the user can now add variables as key work arguments
- additional write function for conflicts file

Known issues:
@manuelma, when I call any of the write_... functions within SpineOpt, it seems to fail writing them (through console after the run, it works fine)

### DIFF
--- a/data/spineopt_template.json
+++ b/data/spineopt_template.json
@@ -115,6 +115,7 @@
         ["model", "write_lodf_file", false, "boolean_value_list", "A boolean flag for whether the LODF values should be written to a results file."],
         ["model", "write_mps_file", null, "write_mps_file_list", "A selector for writing an .mps file of the model."],
         ["model", "write_ptdf_file", false, "boolean_value_list", "A boolean flag for whether the LODF values should be written to a results file."],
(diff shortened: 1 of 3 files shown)


mihlema:
Obviously, this one is breaking things. I'll fix those issues asap

mihlema:
@manuelma I'm about to delete this branch just because it's too outdated by now. 
Still possibly interesting takeaways:
- Do we currently have the option in SpineOpt to report conflicts?
- In the original merge request, functionality is provided that allows the user to have a look at those equations that cause the problem to become infeasible - which has been very very useful for debugging for me.
- As far as I know this is only possible if the model is "direct" , is this still something we support?
- Bonus addition that this branch tried to incorporate: "add_user_variables"  and database parameter "write_math_file" (with somewhat different reporting schema than mps file)

manuelma:
> Do we currently have the option in SpineOpt to report conflicts?
In the original merge request, functionality is provided that allows the user to have a look at those equations that cause the problem to become infeasible - which has been very very useful for debugging for me.
As far as I know this is only possible if the model is "direct" , is this still something we support?

We currently report conflicts for solvers that support it. This works without direct_model too. I think we are good regarding this.

Not sure about write_math_file though... We might want to incorporate that in master?

PR #367, opened 2021-04-15
Hjy documentation

nnhjy:
There is a note in `output.md` that needs some toughts.


Tasqu (approved):
I made a few minor revisions and fixed a few broken references.

PR #372, opened 2021-05-07
Add an entry script to SpineOpt

ghost:
This is what I meant with the entry script to SpineOpt. My confusion earlier was that I thought there would have already been this kind of script for running SpineOpt included.

You can test this by running:

    $ julia run.jl sqlite:///path/to/db.sqlite sqlite:///out.sqlite

See full usage with 

    $ julia run.jl --help

Maybe this is now overlapping a bit with the SpineOptPlugin. Let’s discuss what are the implications, if you feel that this could be included here.


ererkka:
Oops, I realised that it’s not as simple when running on a new machine. I had PyCall set up properly and so it worked well, but without `spinedb_api` it does not work. However, as we know, there is no easy way to set that up automatically/easily.

manuelma:
This is really nice. However, does it work with the plugin? How does the SpineOpt tool spec in the plugin knows where the `run.jl` is located, if it's part of the SpineOpt repo?

I'm also not sure about doing package stuff (instantiate, etc) as part of the 'run'. It feels this could be done separately but I'm not sure.

Issue #376, opened 2021-05-11
Long `connection_lifetime` values are very bad for performance


manuelma:
I have an investment model that spans 10 years, that I solve using Benders. In the operations problem, both `model.roll_forward` and `temporal_block.resolution` are set to 1 week. Thus, each sub-problem has exactly one 'present' time slice and one history time slice. This goes well.

However I tried to set `connection_lifetime` to five years, and this has the effect that the history of each subproblem is now 5 years, that is 5 * 52 time slices of one-week duration. Rolling takes a much longer time, because the model basically needs to sweep all those 261 time indices to shift the variables 'one position to the left'.

It feels like we need a better solution. I don't think all 5 years of history need to have the same resolution as the present?

manuelma:
We discussed this and there are two options:

1. Determine the history length on a per-problem basis. For the operations problem, we don't need to consider the <investment>_lifetime parameter.
2. Add an additional time-slice index to the investment variables to represent the moment of investment (vintage year?). This seems to also solve the issue but it is uncertain, at least to me, how big would be the impact. @DillonJ suggests it can be quite big.

DillonJ:
I have an investments model with unit_investment_lifetime = 364D and a 24hour rolling window. Of course this doesn't make a ton of sense but it results in a traceback:

```
Updating variables...ERROR: LoadError: KeyError: key (unit = EL_GAS_CT_NEW1(I), stochastic_scenario = Deterministic, t = 2029-01-05T00:00~>2030-01-04T00:00) not found
Stacktrace: [1] getindex(::Dict{NamedTuple{(:unit, :stochastic_scenario, :t),Tuple{SpineInterface.Object,SpineInterface.Object,SpineInterface.TimeSlice}},Real}, ::NamedTuple{(:unit, :stochastic_scenario, :t),Tuple{SpineInterface.Object,SpineInterface.Object,SpineInterface.TimeSlice}}) at .\dict.jl:477
 [2] update_variable!(::Model, ::Symbol, ::Function; update_names::Bool) at C:\Workspace\Spine\SpineOpt.jl\src\run_spineopt_sp.jl:162
 [3] update_variable!(::Model, ::Symbol, ::Function) at C:\Workspace\Spine\SpineOpt.jl\src\run_spineopt_sp.jl:148
 [4] update_variables!(::Model) at C:\Workspace\Spine\SpineOpt.jl\src\run_spineopt_sp.jl:169
[...]

PR #378, opened 2021-05-19
Splitting history in short-term and long-term (see also #376)

mihlema:
- the saved history is separated into operational and investment delay parameters (dependent on unit/node/connetion__(investment)_temporal_block etc.)
- @manuelma, could you try it out, if it decreases your computational times? (Also if see any problems/performance killers let me know)

### DIFF
--- a/src/data_structure/temporal_structure.jl
+++ b/src/data_structure/temporal_structure.jl
@@ -198,22 +198,39 @@ function _generate_time_slice!(m::Model)
     window_start = start(window)
     window_end = end_(window)
     window_time_slices = _window_time_slices(instance, window_start, window_end)
-    history_time_slices = Array{TimeSlice,1}()
-    required_history_duration = _required_history_duration(instance)
     window_duration = window_end - window_start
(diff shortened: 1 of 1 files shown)


mihlema:
also related: #376

mihlema:
@manuelma another stale branch, but I'm thinking this one is in somewhat better shape, and might actually be quite useful for SpineOpt. For each variable, we can choose whether we will need the "long"  investment history, or only the "short" operational history. Should drastically reduce look-back history for operational variables.
But maybe we have already done something like this?

manuelma:
> @manuelma another stale branch, but I'm thinking this one is in somewhat better shape, and might actually be quite useful for SpineOpt. For each variable, we can choose whether we will need the "long" investment history, or only the "short" operational history. Should drastically reduce look-back history for operational variables. But maybe we have already done something like this?

Nice one @mihlema I will look into this.

mihlema:
Also related to #606

datejada:
Closing this PR because it was addressed on PR #910

PR #380, opened 2021-06-01
Update to Installation chapter

PekkaSavolainen:
I think this makes the installation section a bit cleaner, up-to-date, and more consistent with Spine Toolbox User Guide. Please comment if there's something I missed or does not make sense.


manuelma:
Thanks very much @PekkaSavolainen 

I think we can make it even more transparent and straightforward. I will give it a shot.

My concerns are the following:
- We mention Conda as if it were some kind of dependency. It's not. SpineOpt only requires Julia and spinetoolbox to run.
- Optional steps are not justified. We should tell the user what they do. Moreover, I think we can put those in a separate section below. I think users want to see shorter instructions.

PekkaSavolainen:
Go ahead @manuelma. I did not change a lot. Just the parts that were clearly inconsistent with Toolbox User Guide.

Issue #384, opened 2021-06-02
Archetypes - initial implementation using templates


mihlema:
As discussed before, our initial step towards supporting archetypes will be to provide `.json` templates that can be imported with all necessary parameter relationships defined. This issue serves as a list of envisioned archetypes and responsibilities. Note that archetypes will also need to be documented in the online documentation #280, in general (how to use) and very briefly specific (what is each archetype for).
Note that I have also added "model" archetypes (in contrast to "system" archetypes, which typically correspond to the physical system)

 ~~General description on how to use archetypes @mihlema~~ (I feel like this should be mentioned in the relevant parts of the documentation; e.g. description of pressure-driven gas transfer) 

System archetypes:
- [x] Units with ramps @mihlema @KristofPhillips95 
    - [x] template
    - [x] documentation
- [x] Units with ramps and reserves @mihlema @KristofPhillips95 
    - [x] template
    - [x] documentation
- [x] Units with unit-commitment constraints @mihlema @KristofPhillips95
    - [x] template
    - [x] documentation 
- [ ] Pressure-driven gas pipeline @mihlema
  - [ ] template
   - [ ] documentation
- [ ] PTDF-based (minimal) network @DillonJ 
    - [ ] template
    - [ ] documentation
- [ ] Node-based DC power flow @mihlema
    - [ ] template
    - [ ] documentation
- [ ] others?

Model archetypes, t.b.d.:
- [x] Basic model with 1 temporal block, 1 stochastic scenario @Tasqu, @manuelma?
    - [x] template
    - [ ] documentation
- [x] Model with simple stochastic tree @Tasqu 
    - [x] template
    - [x] documentation
- [ ] Model with (basic) structure for benders decomposition @DillonJ 
    - [ ] template
    - [ ] documentation
- [ ] others?

I'll make an initial example for one of "my" archetypes, then others can swiftly follow.

Tasqu:
Just to be 100% clear here, should we try and keep the archetypes as "minimal" as possible, meaning that they area meant for importing on top of existing DB:s and each other to serve as "templates" for doing different things in SpineOpt? 

[...]

PR #386, opened 2021-06-04
#384 stochastic structure archetypes

Tasqu:
So I made a first draft of what I think is meant by "archetypes" for the stochastic structures. Essentially, they are just .json database templates that contain the bare minimum entities for a functioning `stochastic_structure` ready to be linked to a `model` and `nodes`/`units`.

- I don't know where we want to actually include these archetypes in the codebase, so currently I just made a folder `data/archetypes/` for the .json files.

- Eventually, these probably need to be explained or documented somewhere. Do we create a new section in SpineOpt documentation for there, or something else? I don't think .json supports any docstrings or the like, so I don't know if we can document the files within themselves.


manuelma:
I actually pushed a `basic_spineopt_template.json` file to the `data` folder in master. I'm being a bit hasty here because I need to organize a workshop so I prefer to take shortcuts. Sorry about that. I feel we can always move stuff around and organize things better once we settle on the desired approach?

Tasqu:
> I actually pushed a basic_spineopt_template.json file to the data folder in master. I'm being a bit hasty here because I need to organize a workshop so I prefer to take shortcuts. Sorry about that. I feel we can always move stuff around and organize things better once we settle on the desired approach?

Yeah, of course. I mean, my `archetypes` folder is completely arbitrarily decided by me atm. We also seem to have an `examples` folder in there, so I'm not sure if the archetypes should be listed under there instead. In any case, we can discuss this.

Tasqu:
FYI: This branch contains my proposal for how to organise the different JSON examples in the SpineOpt repository, in addition to the stochastic structure archetypes and their documentation.

This is not intended to be the final product, but my initial proposal for handling the archetypes and their documentation in the repo.

Issue #418, opened 2021-11-03
How to go about stochastics for implementing `min/max_total_cumulated_flows`


mihlema:
Trying to add cumulated bounds to SpineOpt, we have formulated the following constraint:
```
function add_constraint_total_cumulated_unit_flow!(m::Model, bound,sense,d)
    @fetch unit_flow = m.ext[:variables]
    m.ext[:constraints][bound.name] = Dict(
        (unit=ug, node= ng, stochastic_path = s,t=t ) => sense_constraint(
            m,
            + expr_sum(
                unit_flow[u, n, d, s, t] * duration(t)
                for (u, n, d, s, t) in unit_flow_indices(
                    m;
                    unit=ug,
                    node = ng,
                    direction=d,
                    stochastic_scenario=s);
                    init = 0
            ),
            sense,
            bound(unit = ug, node = ng)
        ) for (ug,ng,s) in constraint_total_cumulated_unit_flow_indices(m,bound,d)
    )
end
```
where bound `sense` can be `>=, <=, =` and bound is either `max, min`. The bound should be over the entire modeling horizon (*Note: There will be another constraint following for the annual cumulative bounds*). The tricky part here seems to be the stochastics - and what exactly we want to achieve.

As indicated above, the `constraint_total_cumulated_unit_flow_indices(m,bound,d)` does  return a tuple of `ug`,`ng` and stochastic_path `s`. The `bound` parameter itself, however doesn't have temporals nor stochastics. For the temporals, this is a logical, as we want to cumulate over the entire model horizon. For the stochastics, this might not be straightforward. There's two things to decide/discuss
## Do we want a scenario-dependent `bound`?
This is the first decision we need to look at. For example, is there a case where we want to have different max/min bounds for different investment cases? I think we might - or at least maybe we shouldn't exclude it ex-ante. In this case, our bound would have another index `stochastic_scenario = s`, where s would correspond to the investment scenario. So how to retrieve this stochastic scenario though? We could use the `unit_investment__stochastic_structure` relationship for this. But how would this then work in problems without an investment decision? 
## Should the `bound` be imposed for each scenario path separately?
[...]

PR #419, opened 2021-11-09
WIP: merging cumulated flows into Master

mihlema:
We've added the functionality and test for the cumulated max/min flows over the entire modelling horizon. Once docs are finalized we can merge this.

This merge request contains:
- cumulated flow bounds (possibly to be extended; see ideas in #418)
- test for those new constraints
- "Bonus" more data structure checks. Mainly on parameters that need to be defined to ensure that model start<=modelend, ramping parameters defined correctly etc.


src/data_structure/check_data_structure.jl
```diff
@@ -296,3 +297,227 @@ function check_rolling_branching()
         end
     end
 end
+
+function check_basic_requirements_parameter_values()
```
manuelma, line 301: I propose we call this just check_parameter_values

manuelma:
Nice job!

PR #425, opened 2022-02-08
Update issue templates

mihlema:
Hi all,
what's your opinion on these issue templates?


mihlema:
@DillonJ thanks for the comment. I have tried to make the `bug-report` template a bit more SpineOpt focused, let me know what your thoughts are.

mihlema:
@jkiviluo is this still something we want to have? Otherwise I'll drop the (stale) merge request.

Tasqu (approved):
I don't see why we wouldn't want to include these if we've got them 👍

PR #427, opened 2022-02-15
Hot fixes bridging gaps

mihlema:
- The previous implementation of `in_gaps` conflicted with the representatvie period implementation, should work now
- also: outputs are now written for all (not only representative) days. this way e.g. for unit flow's we get a nicer visualization of the evolution over the entire year


mihlema:
@manuelma Maybe if you have time you can have a look at this one (just to make sure). The idea is simply to disable the gap bridger if representative period mapping is used.

PR #435, opened 2022-03-02
Mga implementation

mihlema:
This merge request will add a new algorithm to SpineOpt - not for decompositon but modeling-to-generate-alternatives (MGA).
For now this supports that near optimal solution are explored, i.e. investments in units,nodes,and connections can be as different as possible, while the objective function needs to be Within the close range of the original objective function.

Also got rid of lazy include
(Almost ready to merge, but some clean-up still necessary and migration script needs to be adapted. )


mihlema:
I'm struggling a bit with the replacing of parameter values, specifically `spineopt_master` should be renamed `spineopt_benders_master` and `spineopt_operations` should be `spineopt_standard`. I guess at the same time we would also need to update the `model_type_list`?
@manuelma , @DillonJ , do you have a suggestion (see versions/rename_model_types.jl)

src/data_structure/MGA_data.jl
```diff
@@ -0,0 +1,209 @@
+#############################################################################
```
DillonJ, line 1: Generally speaking, all our filenames are in lower case - this filename seems to be departing from the convention

test/data_structure/algorithm_MGA_structure.jl
```diff
@@ -0,0 +1,334 @@
+#############################################################################
```
DillonJ, line 1: Generally, all our filenames are lowercase

src/SpineOpt.jl
```diff
@@ -40,151 +41,138 @@ include("run_spineopt.jl")
...
-	"constraints/constraint_max_node_pressure.jl",
-	"constraints/constraint_min_node_pressure.jl",
-	"constraints/constraint_max_node_voltage_angle.jl",
-	"constraints/constraint_min_node_voltage_angle.jl",
-	"constraints/constraint_node_voltage_angle.jl",
-	"constraints/constraint_connection_unitary_gas_flow.jl",
-	"constraints/constraint_mp_any_invested_cuts.jl",
-]
+include("run_spineopt_sp.jl")
+include("run_spineopt_benders_algorithm.jl")
```
DillonJ, line 45: If we're renaming files, then perhaps we should go for more consistency - perhaps:

```
run_spineopt_standard.jl
run_spineopt_benders.jl
run_spineopt_mga.jl

```

Also, till now we have filenames in lower case only - my preference would be to preserve this.

src/data_structure/MGA_data.jl
mihlema: Thanks, yes that makes sense

DillonJ:
Hi @mihlema this is great stuff. I see you've replied to some of my comments already. I intended only to add comments but at some point hit "start a review" by mistake so sorry about that!

Is the algorithm documented somewhere? Perhaps I missed it?

mihlema:
Thanks @DillonJ. And thanks for the comments already, I'll immediately get to them. Also, I'll add an advanced concept documentation on how to use is/what it does. 
Any idea on how to replace a parameter value list/parameter values in a migration script?

DillonJ:
> Any idea on how to replace a parameter value list/parameter values in a migration script?
Apart from hacking something @manuelma has done for other migrations, I'm not sure - I'll have a look

DillonJ:
regarding renaming paramerer values in a list I think this won't be easy - the values themselves may not have an ID - you might have to add the new one and remove the old one - but then you will have to make sure to update the values of any parameters that use that list value accordingly

DillonJ:
That sounds like a nice solution @manuelma 

Also, we need to check if any parameter definition is using the list value as its default

manuelma:
> That sounds like a nice solution @manuelma
> 
> Also, we need to check if any parameter definition is using the list value as its default

Yes, could be a bit of work but it's doable. We have been wanting to improve parameter_value_list design at DB level, there should be an issue on that somewhere. I think we (I) made some poor choices when first impementing this.

Another door this would open is reusing values from list in different places to avoid repetition. Because at the moment, if two parameter values use the exact same huge time series, we need to write it twice. But if we could just point at values from list, we'd only need to write the time series once in a value list.

DillonJ:
> Another door this would open is reusing values from lists in different places to avoid repetition. Because at the moment, if two parameter values use the exact same huge time series, we need to write it twice. But if we could just point at values from lists, we'd only need to write the time series once in a value list.

Very good point - we're actually getting symbols for free.

Perhaps the values should have their own IDs and then you do a list-value to parameter-list mapping - so a certain list value can be part of multiple value-lists (or none for that matter).

manuelma:
This is the issue where we discuss a redesigning of parameter_value_list https://github.com/Spine-project/Spine-Database-API/issues/151 that I think is a prerequisite for introducing the "from_list" value type. It's not straitghforward but I could check if it can be done quickly.

mihlema:
So everything seems to work here now. - except for the migration process.
I'm not sure if it would suffice to ask the user to load the new template and rename? 

I tried to follow @DillonJ suggestion regarding the renaming, but importing a new parameter value list doesn't seem to work the way I thought it would:
```
###similarly to other migration scripts, we have the new data here:
new_data[:parameter_value_list]
### this throws an error:
run_request(db_url, "import_data", (new_data, ""))

Error: TypeError("get_data_for_import() got an unexpected keyword argument 'parameter_value_list'")
```

src/data_structure/mga_data.jl
```diff
@@ -0,0 +1,219 @@
...
+            iteration
+        )
+        @fetch mga_aux_diff, mga_objective = m.ext[:variables]
+        ub_objective = get!(m.ext[:constraints],:mga_objective_ub,Dict())
+        ub_objective[iteration] = @constraint(
+                m,
+                mga_objective[(model = m.ext[:instance],t=current_window(m))]
+                <= sum(
+                mga_aux_diff[ind...]
+                       for ind in vcat([storages_invested_mga_indices(iteration),connections_invested_mga_indices(iteration),units_invested_mga_indices(iteration)])
```
manuelma, line 102: please indent the `for` at the same level of the previous line as is our style for comprehensions.

src/data_structure/mga_data.jl
```diff
@@ -0,0 +1,219 @@
...
+                   )
+                  + mga_variable_bigM(;ind...)*(1-mga_aux_binary[(ind...,mga_iteration=mga_current_iteration)])
+                  )
+                  d_diff_lb1[(ind...,mga_current_iteration...)] = @constraint(
+                    m,
+                    mga_aux_diff[((ind...,mga_iteration=mga_current_iteration))]
+                    >=
+                    sum(
+                    (variable[_ind]
+                      - mga_results[variable_name][((Base.structdiff(_ind,NamedTuple{(:t,)}(_ind.t))..., mga_iteration=mga_current_iteration))][t0.ref.x][_ind.t.start.x])
```
manuelma, line 174: The call to Base.strucdiff is pretty difficult to read here. Is it just dropping a namedtuple key? Don't we have a function called drop_key somewhere to do that?

src/data_structure/mga_data.jl
```diff
@@ -0,0 +1,219 @@
...
+                           for _ind in variable_indices_function(m; ind...)
+                   )
+                   )
+                   d_diff_lb2[(ind...,mga_current_iteration...)] = @constraint(
+                    m,
+                    mga_aux_diff[((ind...,mga_iteration=mga_current_iteration))]
+                    >=
+                    sum(
+                    - (variable[_ind]
+                      - mga_results[variable_name][((Base.structdiff(_ind,NamedTuple{(:t,)}(_ind.t))..., mga_iteration=mga_current_iteration))][t0.ref.x][_ind.t.start.x])
```
manuelma, line 186: t0.ref.x ? What does it mean and is it possible to make it more transparent?

src/run_spineopt_mga.jl
```diff
@@ -0,0 +1,61 @@
+function rerun_spineopt!(
+    ::Nothing,
+    ::Nothing,
+    m::Model,
```
manuelma, line 4: If this is the mga case, we can name the argument `m_mga`so the reader knows immediately?

src/data_structure/mga_data.jl
```diff
@@ -0,0 +1,219 @@
...
+            d_diff_lb2 = get!(m.ext[:constraints],:mga_diff_lb2,Dict())
+            for ind in mga_indices()
+                d_diff_ub1[(ind...,mga_current_iteration...)] = @constraint(
+                    m,
+                    mga_aux_diff[((ind...,mga_iteration=mga_current_iteration))]
+                    <=
+                    sum(
+                    + (
+                    variable[_ind]
+                     - mga_results[variable_name][((Base.structdiff(_ind,NamedTuple{(:t,)}(_ind.t))..., mga_iteration=mga_current_iteration))][t0.ref.x][_ind.t.start.x]
```
manuelma, line 150: We probably need to break these long lines.

src/SpineOpt.jl
```diff
@@ -40,151 +41,138 @@ include("run_spineopt.jl")
 include("util/docs_utils.jl")
 include("data_structure/migration.jl")
 
-_lazy_include_file_paths = [
```
manuelma, line 43: Since we are getting rid of these we do not need to Base.invokelatest JuMP API anymore.

DillonJ:
@mihlema just checking - does the migration deal with the default value for `model_type`. Currently it is `:spineopt_operations`. I believe it should be changed to `:spineopt_standard` as part of the migration

mihlema:
@DillonJ thanks for the hint. It did work for the cases I toyed around with, but not in the case that no `spineopt_operations` model exists in the db. This is fixed now, thanks!

mihlema:
Hi folks @DillonJ @manuelma 
so I ran into a small problem here yesterday. After addressing your comments and fixing some final things, I was ready to merge. However, tests (passing on my local machine) don't pass on github. I investigated a bit, tried to revert things I did before, but nothing seemed to work. Then I went back the the last commit for which tests passed (a9991717cc79f0a34b359f0b69a368164c2dadd9), and simply checked if this one was still passing (simply created an identical branch, 901c20181639b3133e66f2dd98bf06f2f9fc0d57) just to realize that this one is neither passing anymore. 
Was there possibly a recent change in spinedb-api, or something else that could cause this? The failing test is somewhat obscure (and as said before, the exact same commit was passing tests before)
Leaving a chunk from the tests here:
```
  caused by: MethodError: Cannot `convert` an object of type SpineInterface.IdentityCall{Nothing, Float64} to an object of type SpineInterface.OperatorCall{typeof(+)}
  Closest candidates are:
    convert(::Type{T}, ::T) where T at essentials.jl:205
  Stacktrace:
    [1] setindex!(h::OrderedCollections.OrderedDict{VariableRef, SpineInterface.OperatorCall{typeof(+)}}, v0::SpineInterface.IdentityCall{Nothing, Float64}, key0::VariableRef)
      @ OrderedCollections ~/.julia/packages/OrderedCollections/PRayh/src/ordered_dict.jl:325
    [2] _map_aff(f::typeof(MutableArithmetics.copy_if_mutable), constant::SpineInterface.OperatorCall{typeof(+)}, a::GenericAffExpr{Call, VariableRef})
      @ JuMP ~/.julia/packages/JuMP/R2Knd/src/aff_expr.jl:318
    [3] map_coefficients(f::typeof(MutableArithmetics.copy_if_mutable), a::GenericAffExpr{Call, VariableRef})
      @ JuMP ~/.julia/packages/JuMP/R2Knd/src/aff_expr.jl:313
    [4] mutable_copy
      @ ~/.julia/packages/JuMP/R2Knd/src/mutable_arithmetics.jl:146 [inlined]
    [5] copy_if_mutable_fallback
      @ ~/.julia/packages/MutableArithmetics/BOohi/src/interface.jl:264 [inlined]
    [6] copy_if_mutable
      @ ~/.julia/packages/MutableArithmetics/BOohi/src/interface.jl:275 [inlined]
    [7] operate
      @ ~/.julia/packages/MutableArithmetics/BOohi/src/interface.jl:189 [inlined]
    [8] operate
      @ ~/.julia/packages/MutableArithmetics/BOohi/src/rewrite.jl:34 [inlined]
    [9] operate_fallback!!
      @ ~/.julia/packages/MutableArithmetics/BOohi/src/interface.jl:549 [inlined]
   [10] operate!!(op::typeof(MutableArithmetics.add_mul), x::MutableArithmetics.Zero, args::GenericAffExpr{Call, VariableRef})
      @ MutableArithmetics ~/.julia/packages/MutableArithmetics/BOohi/src/rewrite.jl:88
   [11] macro expansion
      @ ~/.julia/packages/MutableArithmetics/BOohi/src/rewrite.jl:288 [inlined]
   [12] macro expansion
      @ ~/.julia/packages/JuMP/R2Knd/src/macros.jl:815 [inlined]
   [13] #1300
      @ ~/work/SpineOpt.jl/SpineOpt.jl/src/constraints/constraint_node_injection.jl:30 [inlined]
   [14] (::SpineOpt.var"#1300#1311"{Model, SpineInterface._DateTimeRef, Dict{NamedTuple{(:unit, :node, :direction, :stochastic_scenario, :t), Tuple{Object, Object, Object, Object, TimeSlice}}, VariableRef}, Dict{Any, Any}, Dict{NamedTuple{(:node, :stochastic_scenario, :t), Tuple{Object, Object, TimeSlice}}, VariableRef}})(::NamedTuple{(:node, :stochastic_path, :t_before, :t_after), Tuple{Object, Vector{Object}, TimeSlice, TimeSlice}})
      @ SpineOpt ./none:0
   [15] iterate
      @ ./generator.jl:47 [inlined]
   [16] grow_to!(dest::Dict{Any, Any}, itr::Base.Generator{Vector{NamedTuple{(:node, :stochastic_path, :t_before, :t_after), Tuple{Object, Vector{Object}, TimeSlice, TimeSlice}}}, SpineOpt.var"#1300#1311"{Model, SpineInterface._DateTimeRef, Dict{NamedTuple{(:unit, :node, :direction, :stochastic_scenario, :t), Tuple{Object, Object, Object, Object, TimeSlice}}, VariableRef}, Dict{Any, Any}, Dict{NamedTuple{(:node, :stochastic_scenario, :t), Tuple{Object, Object, TimeSlice}}, VariableRef}}})
   ```

manuelma:
> Edit: I checked and realized that also SpineOpt master has the same issue, possibly related to julia 1.6/SpineInterface/JuMP/spinedb-api update maybe?

Could be a JuMP issue. Tests were passing with 0.22.3 and fail with 0.23.1. We had JuMP issue pinned to 0.21.x for some reason and removed it recently, maybe that's why we had it pinned... Will try to fix.

manuelma:
I confirm the issue is with JuMP 0.23.1. @DillonJ what was the reason to unpin JuMP version?

manuelma:
Whatever the reason for unpinning JuMP version, I think it's the right call anyways and we absolutely need to support latest JuMP, so I'm working on a fix from SpineInterface.

mihlema:
Thanks a million @manuelma for taking this up!

DillonJ:
>I confirm the issue is with JuMP 0.23.1. @DillonJ what was the reason to unpin JuMP version?

Yeah - I unpinned the version to install the latest version of HiGHS - I didn't intend to push that change though, so sorry about that. But I agree - we should be supporting the latest version of JuMP

manuelma:
Seems like things work again in master...

mihlema:
Thanks @manuelma for the fix in SpineInterface. Now also this branch is passing tests again.

Issue #326, opened 2021-02-19
Economic representation of investments


mihlema:
In the current implementation, the investment costs appear as a whole at the time of investment.
The investment costs are determined as follows (simplified equation):
```
   investment_costs[s,t] = sum( units_invested[u, s, t] *  unit_investment_cost[(unit=u, t=t)] *
            for (u, s, t) in units_invested_available_indices(m; unit=indices(unit_investment_cost)) 
    )
```
Usually, the investments will not take place in one timestep, but over multiple timesteps (multiple-years). Our current implementation does not properly account for the time value of money. To account for multi-year investments, we might want to  convert our objective function to the net-present value of all (annualized) investment and operational costs. Thoughts @DillonJ @jkiviluo ?

Also, we don't take into account any salvage value at the end of the optimization.

TimMertens:
This document summarizes how I included the economic representation of investment in my planning models up till now. This could serve as a basis for a discussion on how we intend to tackle this issue within Spine. The document is roughly based on the approaches taken within TIMES and the LYSUM invest model. (https://www.mech.kuleuven.be/en/tme/research/energy_environment/Pdf/WP-en2018-07)

[Spine_invest.pdf](https://github.com/Spine-project/SpineOpt.jl/files/6350472/Spine_invest.pdf)

clizbe:
Closing due to stale status.

Issue #325, opened 2021-02-19
Investments with technology degredation


mihlema:
This issue addresses the missing functionality of mimicking the degradation process of investments. Currently, we only account for an economic lifetime of investments, i.e. how long until a unit has paid for itself (@DillonJ please correct me if this is incorrect). 

The technical lifetime is not represented, meaning that a unit once installed can be used forever. Moreover, the performance of a unit does not change over time. To account for aging processes we should consider taking this into account.
Missing functionality in this regard is hence:
- [ ] decommissioning of units after their technical lifetime
- [ ] continuous degradation of the efficiency over time
- [ ] (mothballing of technology - not now)

The first point should be quite straight forward, similar to the current implementation of `add_constraint_unit_lifetime!` (which should be renamed to economic lifetime), but would enforce an upper bound on `units_invested_available`.
But we should also be careful about units that were available at the start of the optimization,i.e. `number_of_units`, and made sure we are able to decomission them.

The second point is slightly more tricky. One option could be to have a constant factor, giving the efficiency decrease percentage per year, e.g. 1%/year <sup>1</sup>. This degradation factor would be multiplied with the `fix_ratio_out_in`, `max_ratio_out_in` (per year after installation). For a non-constant degradation factor e.g. `[year1: 1%, year2: 2% ...]`, I think we would need a mapping parameter of the form `duration -> value`, but we don't have support for this from SpineInterface. 

Edit: We do have support from SpineInterface for this, using the mapping parameter. I wonder if this mapping would work in combination with stochastic scenarios?
<sup>1</sup> Arguable [year] unit, but that's another discussion: #82

PR #448, opened 2022-03-25
WIP: Merge economic representation of investments

mihlema:
With this merge request, we will allow for multi-period investments in SpineOpt while taking into account the value of time. All costs are discounted accordingly. Also, the user is now able to define lead times, technical- and economical lifetimes, decommissioning times for units, nodes, and connections. Units and nodes can also be (de-)mothballed, or retired early. Finally, we differentiate between `investment_variables` and `investment_variables_vintage`, which will also allow us to differentiate technology characteristics by vintage year, and to include degradation long term.

Before the merge can be executed the following elements still need to be addressed:

- [ ] **Restore functionality**
- [ ] Fix Benders decomposition accordingly
- [ ] Add all investment variables to the spineopt_template
- [ ] Double check variable creation
- [ ] **Document new functionality**
- [ ] Document new economic parameters (technology discount rate, discount rate, discount duration, lead time, mothballing costs, capacity transfer factor, conversion to discounted annuities etc.)
- [ ] Document new investment variables (investment state, investment available, early decommissioned, decommissioned, de-mothballed ...
- [ ] Document new investment constraints (investment state, investment available, early decommissioned, decommissioned, de-mothballed ...
- [ ] **Write tests for all new functionality**
- [ ] For economic parameters
- [ ] For new constraints
- [ ] Tests should also cover e.g. stochastic scenario dependent leadtimes, and rolling horizon
- [ ] Double check scenario implementation for parameters -> possibly add model dimension
- [ ] **Write migration script**
- [ ] New parameters, but also renaming -> unit_lifetime -> unit_economic/technical_lifetime
- [ ] **Define reasonable defaults**
- [ ] By default, the user should not need to define anything, e.g. to run an operational model
- [ ] Make sure that economic structure doesn't need to be generated if problem is operational only
- [ ] Should univested__invested_available be fixed to 0 at the start or rather units_invested if not defined? (one follow the other). Also units_invested_available, if defined would fail without corresponding investment variable?
- [ ] **Data structure check**
- [ ] Economic lifetime smaller equal technical lifetime
- [ ] If only *one* lifetime exists, assume they are equivalent + warning
- [ ] Warn if only decommissioning time is given but not the decommissioning costs
- [ ] Error if discount rate is larger than 1
- [ ] If lead_time `nothing` should default to `0Years`
- [ ] If we want to disallow Benders + Milestone for the moment -> throw error accordingly


mihlema:
This one will first require an in-depth discussion on what the best/desired way is to incorporate vintage and economic discounting into SpineOpt in the future. Then we can see whether there is still value in moving forward with this branch. Related issues are #325 #326 #601 . I believe we also had a discussion on this at some point, but I can't find it back.

mihlema:
#476 is also related to this

datejada:
The main takeaways from the hackathon in the WP4 (MOPO project):

- The stochastic scenario vintage in the current branch does not support if the `lifetime` changes among scenarios (maybe the WACC, too). Add a caveat in the documentation.

- (fix) `unit_discounted_durantion` -> It doesn't work for the ones being invested (but it works for those not being supported).

- Another caveat -> it is not supported when using Benders' decomposition (The combination Milestone years + Benders' decomposition is not supported). If no milestone years are selected, then you can use the decomposition.

- We need to double-check the variable creation for the vintage variables according to the new structure.

- keyword in the code => `history_long`(good for speed up the code). This parameter will help differentiate between variables needing history in the constraints and those not. We need a separate issue. Double-check how much history triggers the lifetime parameter. This problem has a separate issue; see #606 

- The final version should allow to specify two options for the user: The existing one and the milestone year option (refactor the code to have that)

- Include tests (must), documentation, migration script (must), and tutorial

- Rolling horizon is not supported (future improvement)

- Profile both branches with the exact case study (double-check the existing tutorials)

- What if the investment is at half of the year? (maybe, it is not supported)

- More details in #476

gnawin:
Hi @mihlema, we investigated one problem that was left from the hackathon in Leuven.
> - (fix) `unit_discounted_durantion` -> It doesn't work for the ones being invested (but it works for those not being supported).

We found the cause of this error, and proposed a change to resolve it. But we don't yet if the change is correct or not, because it is more related to the intention of your initial code. Maybe you can see more easily what should be changed from there.

Here's a screenshot of the story:

Capture

The yellow boxes show the cause of the initial error and our proposed changes. The problem occurs because for the wind farm, there is only one stochastic scenario, so the function `_find_children` returns empty. We propose to first check whether this is empty or not, if it's empty, use the root scenario instead. 

The orange line and arrow show a new error. When we take the root scenario, there's something wrong with the `t.block`. 

Since we are not fully aware of the intention of this stochastic part, could you have a look and tell us what to do? 

Thank you!

datejada:
List of existing tests (marked if it passes):

- [ ] data_structure/check_economic_structure.jl
- [ ] data_structure/migration.jl
- [ ] data_structure/check_data_structure.jl
- [ ] data_structure/preprocess_data_structure.jl
- [ ] data_structure/temporal_structure.jl: it fails due to the long/short history
- [ ] data_structure/stochastic_structure.jl
- [ ] data_structure/algorithm_mga_structure.jl: all tests failing
- [ ] data_structure/postprocess_results.jl
- [ ] constraints/constraint_unit.jl: 19 tests failing 
- [ ] constraints/constraint_node.jl: 11 tests failing
- [ ] constraints/constraint_connection.jl: 22 tests failing 
- [ ] constraints/constraint_user_constraint.jl: 6 tests failing
- [ ] constraints/constraint_investment_group.jl: all tests failing
- [ ] objective/objective.jl: 2 tests failing
- [ ] util/misc.jl
- [ ] run_spineopt.jl
- [ ] run_spineopt_benders.jl: all tests failing
- [ ] run_examples.jl: all tests failing

gnawin:
List of new tests in data_structure/check_economic_structure.jl (mark checked if tests are passing)

Already added
- [x] "test discounted duration - using milestone years, w/o inv. blocks"
- [x] "test discounted duration - w/o using milestone years"
- [ ] "test investment costs, salvage fraction, decommissioning"
- [x] "test investment cost scaling"
- [ ] "test technological discount factor"

To be added
- [ ] "test capacity transfer factor"

datejada:
Add new tests for the following:

- [ ] objective/unit_fixed_om_costs.jl
- [ ] objective/storage_fixed_om_costs.jl
- [ ] objective/connection_fixed_om_costs.jl
- [ ] objective/unit_decommissioning_costs.jl
- [ ] objective/unit_mothballing_costs.jl
- [ ] objective/connection_decommissioning_costs.jl
- [ ] objective/storage_decommissioning_costs.jl
- [ ] objective/storage_mothballing_costs.jl
- [ ] constraints/constraint_units_decommissioned_vintage.jl
- [ ] constraints/constraint_units_decommissioned.jl
- [ ] constraints/constraint_units_invested_available_bound.jl
- [ ] constraints/constraint_units_invested_available_vintage.jl
- [ ] constraints/constraint_units_invested_available.jl
- [ ] constraints/constraint_units_invested_state_vintage.jl
- [ ] constraints/constraint_units_invested_state.jl
- [ ] constraints/constraint_units_mothballed_state_vintage.jl
- [ ] constraints/constraint_connections_decommissioned_vintage.jl
- [ ] constraints/constraint_connections_decommissioned.jl
- [ ] constraints/constraint_connections_invested_available_bound.jl
- [ ] constraints/constraint_connections_invested_available_vintage.jl
- [ ] constraints/constraint_connections_invested_available.jl
- [ ] constraints/constraint_storages_decommissioned_vintage.jl
- [ ] constraints/constraint_storages_decommissioned.jl
- [ ] constraints/constraint_storages_invested_available_bound.jl
- [ ] constraints/constraint_storages_invested_available_vintage.jl
- [ ] constraints/constraint_storages_invested_available.jl
- [ ] constraints/constraint_storages_invested_state_vintage.jl
- [ ] constraints/constraint_storages_invested_state.jl
- [ ] constraints/constraint_storages_mothballed_state_vintage.jl

src/constraints/constraint_connections_decommissioned.jl
```diff
@@ -0,0 +1,54 @@
...
+                for (c, s, t_v, t) in connections_invested_available_vintage_indices(
+                            m;
+                            connection=c,
+                            stochastic_scenario=s,
+                            t=t,
+                            t_vintage = to_time_slice(
+                                m;
+                                t=TimeSlice(
+                                    isnothing(connection_investment_tech_lifetime(connection=c)) ?
+                                    t0.ref.x #FIXME: needs to look back in history
```
manuelma, line 44: does this work? the field ref has been renamed I think. For this exact reason I would try and avoid accessing fields, they are subject to renaming. Maybe define a function that returns the field, that's better style.

src/constraints/constraint_storage_lifetime.jl
```diff
@@ -33,10 +33,22 @@ function add_constraint_storage_lifetime!(m::Model)
...
             )
-            >=
-            sum(
-                storages_invested[n, s_past, t_past]
-                for (n, s_past, t_past) in _past_storages_invested_available_indices(m, n, s, t)
+            >= #FIXME: >= for econ lifetime, <= for tech lifetime
+            + sum(
+                + storages_invested[n, s_past, t_past]
+                * storage_capacity_transfer_factor[(node=n, stochastic_structure__stochastic_scenario=s_past,vintage_t=first(t_past.start),t=t)]
+                for (n, s_past, t_past) in storages_invested_available_indices(
```
manuelma, line 40: How come stochastic_structure__stochastic_scenario is one of the dimensions? That doesn't sound right. Also, as said above, accessing a field (t_past.start) is asking for trouble, better use start(t_past)

src/constraints/constraint_connections_invested_available_vintage.jl
```diff
@@ -0,0 +1,104 @@
...
+                            m;
+                            connection=c,
+                            stochastic_scenario=s,
+                            t=t_v,
+                            )
+                ; init=0
+                )
+        ) for (c, s, t_v, t) in constraint_connections_invested_available_vintage_indices(m)
+    )
+end
```
manuelma, line 68: Would be nice to improve the style in this function, fix the indentation, remove unnecessary line breaks etc...

src/constraints/constraint_storages_decommissioned_vintage.jl
```diff
@@ -0,0 +1,76 @@
...
+        ) for (n, s, t_v, t_after) in constraint_storages_decommissioned_vintage_indices(m)
+    )
+end
+
+function constraint_storages_decommissioned_vintage_indices(m::Model)
+    unique(
+        (node=n, stochastic_path=path, t_vintage=t_v, t=t)
+        for (n, s, t_v, t) in storages_invested_available_vintage_indices(m)
+        for path in active_stochastic_paths(_constraint_storages_decommissioned_vintage_indices(m, n, s, t))
+    )
```
manuelma, line 53: This indexing style leads to exponential lookup. We have better techniques (t_lowest_resolution_paths) currently. Not sure if they are applicable but the way it is now, can be very bad for performance.

src/data_structure/economic_structure.jl
```diff
@@ -0,0 +1,491 @@
...
+function generate_economic_structure!(m::Model;log_level=3)
+    for (obj, name) in [(unit,:unit),(node,:storage),(connection,:connection)]
+        @timelog log_level 3 "- [Generated discounted durations for $(obj)s]" generate_discount_timeslice_duration!(m::Model,obj, name)
+    end
+    !isempty([
+        model__default_investment_temporal_block()...,
+        node__investment_temporal_block()...,
+        unit__investment_temporal_block()...,
+        connection__investment_temporal_block()...
+    ]) || return
```
manuelma, line 34: This splatting is not very fast, maybe use vcat - or repace the `...,` by just `;`

src/data_structure/economic_structure.jl
```diff
@@ -0,0 +1,491 @@
...
+                            :stochastic_scenario=>s,
+                            :t => Iterators.flatten((history_time_slice(m), time_slice(m)))
+                            )...
+                        )
+                    LT = lead_time(;Dict(obj_cls.name=>id,:stochastic_scenario=>s,:t=>vintage_t)...)
+                    if isnothing(LT)
+                        LT = Year(0)
+                        #NOTE: In case LT is `none`, we will assume a duration of `0 Years`
+                    end
+                    TLIFE = tech_lifetime(;Dict(obj_cls.name=>id,:stochastic_scenario=>s,:t=>vintage_t)...)
```
manuelma, line 83: Uppercase variable name?

src/data_structure/economic_structure.jl
```diff
@@ -0,0 +1,491 @@
...
+
+"""
+function generate_conversion_to_discounted_annuities!(m::Model, obj_cls::ObjectClass, obj_name::Symbol)
+    instance = m.ext[:spineopt].instance
+    discnt_year = discount_year(model=instance)
+    conversion_to_discounted_annuities = Dict()
+    investment_indices = eval(Symbol("$(obj_name)s_invested_available_indices"))
+    lead_time = eval(Symbol("$(obj_name)_lead_time"))
+    econ_lifetime = eval(Symbol("$(obj_name)_investment_econ_lifetime"))
+    param_name = Symbol("$(obj_name)_conversion_to_discounted_annuities")
```
manuelma, line 156: This eval and Symbol interpolation are not the most readable, would you consider passing all the arguments explicitely?

manuelma:
Very nice job so far! I added a few comments here and there.

So the change is, to say the least, substantial. How do we gain enough confidence to merge this into master? It would be nice to see a more detailed discussion of the code design, to help one understand the code. The fille economic_structure.jl seems very complicated too.

Any chance we can split this into smaller, more managle PRs? At the moment it feels very overwhelming.

What do you think @datejada @g-moralesespana @jkiviluo @gnawin ?

datejada:
Hi @manuelma, thanks for the comments; we really need a deep review and understanding of the changes. So, we need more discussion about the changes and refactoring to improve readability in the new module. We will gain confidence in the proposed changes if we fix the tests. That's why I advocate to work on that part first 😉

manuelma:
> Hi @manuelma, thanks for the comments; we really need a deep review and understanding of the changes. So, we need more discussion about the changes and refactoring to improve readability in the new module. We will gain confidence in the proposed changes if we fix the tests. That's why I advocate to work on that part first 😉

Sounds good @datejada - one thing though is, most of the tests validate that the model generates a certain constraint expression. If you change the constraint in the source, you also need to change it in the tests. However this doesn't test that the **results** of the model are correct or, more accurately, that the results for the a certain dataset are preserved after a change in constraint source code. This is a flaw of our test-suite.

So what we'd need to do in my opinion is expand our test suite with tests that check the results for a good number of representative investment models **under the current design**. If these tests aren't broken after changing the design, I would feel much more comfortable. But then again that can be a lot of work. I don't know how much time you can allocate to this...

Another thought is, this is your first PR and it's hundreds if not thousands line of code. Wouldn't you feel more comfortable starting with something more bounded? One suggestion might be to try and fix that issue with the history (I can give you some pointers, maybe we can even have a chat) so that you get more confidence and get to know some of the internal mechanics much better. That's just an idea of course.

datejada:
> > Hi @manuelma, thanks for the comments; we really need a deep review and understanding of the changes. So, we need more discussion about the changes and refactoring to improve readability in the new module. We will gain confidence in the proposed changes if we fix the tests. That's why I advocate to work on that part first 😉
> 
> Sounds good @datejada - one thing though is, most of the tests validate that the model generates a certain constraint expression. If you change the constraint in the source, you also need to change it in the tests. However this doesn't test that the **results** of the model are correct or, more accurately, that the results for the a certain dataset are preserved after a change in constraint source code. This is a flaw of our test-suite.
> 
> So what we'd need to do in my opinion is expand our test suite with tests that check the results for a good number of representative investment models **under the current design**. If these tests aren't broken after changing the design, I would feel much more comfortable. But then again that can be a lot of work. I don't know how much time you can allocate to this...
> 
> Another thought is, this is your first PR and it's hundreds if not thousands of lines of code. Wouldn't you feel more comfortable starting with something more bounded? One suggestion might be to try and fix that issue with the history (I can give you some pointers, maybe we can even have a chat) so that you gain more confidence and get to know some of the internal mechanics much better. That's just an idea of course.

Hi @manuelma, @gnawin and I were discussing and pondering the options, and we agree that we should split the changes in more controlled PRs. So, we have created an Epic issue #908 with the guideline of changes (based on all the information in this PR) and from there, we can create more minor issues to merge one by one. The strategy remains the same. Reuse all the work from @mihlema, but change the code by smaller PRs.

src/data_structure/economic_structure.jl
gnawin: @manuelma thank you. I tried something to explicitly define the variables/parameters (https://github.com/spine-tools/SpineOpt.jl/pull/929/commits/992f1853ad1ec8cfb797bf179eb3efc9a9316c04), do you think this seems fine? Or do you suggest something different? Once we agree on this, I'll refactor the rest.

src/data_structure/economic_structure.jl
```diff
@@ -0,0 +1,491 @@
...
+                        end
+                        capacity_transfer_factor[(id, vintage_t.start.x, t.start.x)] =  parameter_value(val)
+                        push!(timeseries_val,val)
+                        push!(timeseries_ind,t_start)
+                    end
+                    push!(map_indices,vintage_t_start)
+                    push!(timeseries_array,TimeSeries(timeseries_ind,timeseries_val,false,false))
+                end
+                push!(map_stoch_indices,s)
+                push!(map_inner, SpineInterface.Map(map_indices,timeseries_array))
```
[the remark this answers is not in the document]
gnawin: I'll leave this for later after we agree on the major part 😄

datejada:
Closed in favor of #929

PR #479, opened 2022-08-22
Bug fix: Include correct stochastic path for this constraint

mihlema:
Disclaimer:
Unfortunately, I am a bit pressed for time and don't have time for testing myself. Ideally we would add unit tests e.g. for two-stage stochastic optimization. Current unit tests are passing though. The previous implementation is bugged (see below), so as soon as someone has time to check this, we should merge this quickly.
----
Previously, the investments variable `units_invested_available` would receive the stochastic scenario of the `units_on` variable. However, as investment and operational stochastics are decoupled, different stochastics might be active at different points in time. Now, as we pass all potential paths to the constraint function, we should be able to do this correctly.


Tasqu (approved):
As far as I can tell, this should indeed fix the immediate issue.

PR #481, opened 2022-08-23
WIP: update user constraint generation

mihlema:
- removed `node_state_indices ` (as this resolution is always a subset of `node_injection_indices`
- removing to_node/from_node and replace with unit_flow indices (toDO: add this to preprocessing)
- TODO: set-template values to `null`


mihlema:
Tests are passing now (for julia 1.6, as julia 1.2 seems no longer supported because of @spawn).

datejada:
The aim of this PR is to fix to ensure that resolution is the one that is expected.

Check the current state in the main branch and see if it makes sense for the changes in #1119

Tasqu:
Closing this as stale.

Issue #483, opened 2022-08-23
Passing stochastic paths to parameter calls, e.g. in constraint_connection_flow_capacity.jl


mihlema:
I'm looking at the flow capacity constraint over here, in particular line 50 ff:
```
            - connection_capacity[(connection=conn, node=ng, direction=d, stochastic_scenario=s, analysis_time=t0, t=t)]
            * connection_availability_factor[(connection=conn, stochastic_scenario=s, analysis_time=t0, t=t)]
            * connection_conv_cap_to_flow[
                (connection=conn, node=ng, direction=d, stochastic_scenario=s, analysis_time=t0, t=t),
            ]
```
the `s` that gets passed to these parameters seems to be a stochastic path, rather than a stochastic scenario. I think we'd need to fix this in the future. The problem that I see here is that the summed connection\_flows can have different stochastic scenarios; so which scenario should define e.g. the `connection_capacity` for the entire group (as the parameter can be stochastic too). 
Food for thought I guess. A similar issue is observed in `constraint_total_cumulated_unit_flow_bounds.jl`.

mihlema:
Similarly, for the `fix_ratio` etc. constraints, a similar problem could occur, but here we avoid the problem by associating the parameter with the stochastic scenario of the right-hand side variable.

mihlema:
I'll try to kick-off the conversation here. The tricky part is that we would need to decide, which scenario would take precedence in this situation. For the time slices, we user the lowest resolution as the defining time slice. But I don't think it's quite as straightforward for the stochastics. 

We could have a convention that the constraint will be generated on the most "parent" of the involved scenarios as a start, and see if it suits our purposes. Any thoughts @Spine-project/spineopt-developers ?

Tasqu:
Huh. I'm not even sure how SpineInterface.jl handles a parameter call with a stochastic path instead of a stochastic scenario. I'm guessing it just throws a key error or something?

My intuition would be to try and tie all the parameters to the stochastic indexing of whichever variable they operate upon. E.g. here
```julia
- connection_capacity[(connection=conn, node=ng, direction=d, stochastic_scenario=s, analysis_time=t0, t=t)]
        * connection_availability_factor[(connection=conn, stochastic_scenario=s, analysis_time=t0, t=t)]
        * connection_conv_cap_to_flow[
            (connection=conn, node=ng, direction=d, stochastic_scenario=s, analysis_time=t0, t=t),
        ]
[...]

PR #488, opened 2022-09-21
Bug fix: restore group capacity functionality

mihlema:
This bug was introduced in commit 1a14541933fa0d58189ffb3c5a662dac97e15835 after a series of other adaption. Mainly we summed up the unit_capacity for each unit_flow, while it should be the sum of unit_flows that is constrained by one shared unit_capacity parameter. Bottom line we now retrieve the stochastic scenario for the unit_capacity parameter from the units_on variable again, to restore functionlaity. #483 discusses the problem for other cases

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -22,6 +22,7 @@ jobs:
           - windows-latest
         arch:
           - x64
+        python-version: '3.9'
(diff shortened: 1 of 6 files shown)


mihlema:
Restore group unit_flow functionality and fix PyCall issue for now

Issue #487, opened 2022-09-20
Drop support for julia1.2


mihlema:
I believe we're not supporting julia 1.2 anymore, as we're using functionality (related to the Distributed.jl package?) that is requiring a higher version of julia. This is also why our tests are currently not passing. I think we can move forward at end support for julia 1.2? ( as we're not supporting it anymore anyways?) @manuelma

manuelma:
Done

PR #498, opened 2022-10-20
Update ci.yml

mihlema:
- removing julia 1.2 from workflow operation


mihlema:
Related to #487

manuelma:
So I managed to miss this and just push the exact same changes to master - sorry about that...

PR #506, opened 2022-11-25
Implement MILP representation for piecewise linear unit flow segments

nnhjy:
This feature is implemented as per Discussion #489 to capture the increasing efficiencies problem associated with `unit_flow_op`s. The implementation is built on @mihlema's original design (check the document [here](https://www.overleaf.com/read/dfvxcnhphcgw)).

Currently, MILP models realise the functionality correctly, whereas LP models fail in some cases. Examples can be found in [this DB](https://github.com/nnhjy/Spine-Samples/blob/35355affa6a0eb84b74942876979f88e82819a62/SpineDemo/.spinetoolbox/items/spineopt_demo_db_-_piecewise_unit_flow/SpineOptDB_piecewise-unit-flow.sqlite).

The docs are also renewed accordingly but may need recompiling.


mihlema:
Great to see this shaping, I have converted it to draft for now, but let me know whenever this will be ready for review!

mihlema:
Just for reference, [this](https://www.overleaf.com/read/dfvxcnhphcgw) was my original draft document on overleaf, maybe still good for reference in the future.

DillonJ:
This looks great.

As I mention in https://github.com/Spine-project/SpineOpt.jl/discussions/489 there are significant performance advantages to avoiding use of additional integer variables in the case of convex piecewise linear functions, e.g. when heat rates are monotonically increasing. This was the assumption behind the original implementation. In this implementation, can we preserve this functionality and keep the significant performance benefits? I.e. the user should be able to choose whether to include aditional integers that enforce the ordering of operating point loading, or not and we shouldn't impose additional integer variables in cases where they are not needed (i.e. the convex case, monotonically increasing heat rates).

nnhjy:
I built unit tests for all added/revised constraints. The `pw_heatrate` functionality should be preserved. And no variables will be enforced to be `integer` by default. Details can be found at https://github.com/Spine-project/SpineOpt.jl/discussions/489#discussioncomment-4264375.

manuelma:
This honestly looks great but it introduces an additional variable that is not needed if the efficiency curve (so to say) is monotonically decreasing. We could identify whether that's the case and then fallback to the previous implementation, but ideally I would like to do this as gracefully as possible. I might come up with some ideas very soon. In the meantime I think we can hold on this, if that's ok with you?

src/constraints/constraint_operating_point_bounds.jl
```diff
@@ -20,29 +20,18 @@
...
-            * unit_conv_cap_to_flow[(unit=u, node=n, direction=d, stochastic_scenario=s, analysis_time=t0, t=t)]
-            # TODO: extend to investment functionality ? (is that even possible)
-        ) for (u, n, d) in indices(unit_capacity)
-        for (u, n, d, op, s, t) in unit_flow_op_indices(m; unit=u, node=n, direction=d)
+            unit_flow_op_active[u, n, d, op, s, t]
+            <= 
+            units_on[u, s, t]
+        )
+        for (u, s, t) in units_on_indices(m)
+        for (u, n, d, op, s, t) in unit_flow_op_indices(m; unit=u, stochastic_scenario=s, t=t)
```
mihlema, line 35: The stochastic path of this constraint would break as soon as units on and unit flow would have different stochastic scenarios. The only solution to this at the moment would probably be to use a designated ` constraint_operating_point_bound_indices` function and provide stochastic paths for both variables

mihlema: One thing to keep in mind is how we would constrain e.g. different fuels in these settings. I guess the only way at this point would be to model this via 2 units with a designated user constraints. Alternatively we could also think of a constraint in the form of (draft example):
```julia
sum(unit_flow_op_active(u,n_i,d_i,op,s,t) for (u,n_i,d_i,op,s,t) in
unit_flow_op_indices(unit=u,op=op,s=s,t=t)) <= 1 forall (u,op,s,t)
```
This could be an optional constraint to have a binary choice of which fuels and corresponding heat rates are used.

src/constraints/constraint_unit_flow_op_bounds.jl
```diff
@@ -0,0 +1,49 @@
...
+                - (
+                    (op > 1) ?
+                    operating_points[(unit=u, node=n, direction=d, stochastic_scenario=s, analysis_time=t0, i=op - 1)] :
+                    0
+                )
+            )
+            * unit_capacity[(unit=u, node=n, direction=d, stochastic_scenario=s, analysis_time=t0, t=t)]
+            * unit_flow_op_active[u, n, d, op, s, t]
+            * unit_conv_cap_to_flow[(unit=u, node=n, direction=d, stochastic_scenario=s, analysis_time=t0, t=t)]
+            # TODO: extend to investment functionality ? (is that even possible)
```
mihlema, line 45: Investment functionality should already be covered as 
`unit_flow_op_active <= units_on ,= units_available <= number units + units_invested`

src/constraints/constraint_operating_point_bounds.jl
mihlema: Note that this would only work for binary online_variable_types

mihlema (approved):
Hi @nnhjy, finally had time to have a look at this. Looks all pretty good to me.
My only concern is with regard to the stochastic structure, maybe worth chatting about during our next meeting, but I think it's okay to merge if this is commented in the code. We can talk about it during our next meeting as well.

src/constraints/constraint_operating_point_bounds.jl
nnhjy: Does the `constraint_unit_flow_capacity_indices` provide the needed indices?

manuelma:
> Before merge we need to ensure that the simplified version works (when heat rates are monotonically increasing), probably with a method parameter? Or rename the original parameter to strictly_decreasing_heat_Rate (but that sounds cumbersome).Further we will need a migration script to update existing databases accordingly

@mihlema I think we need to be careful not to introduce unnecessary variables. This new formulation is pretty great but it does so. I suggested to use a method parameter too to let the user choose the treatment of segments. I think it's very important not to merge this PR before doing that.

mihlema:
I agree @manuelma , it's easy to change this now - once merged it's going to drag along.

src/constraints/constraint_operating_point_bounds.jl
nnhjy: I tested it a bit by giving `unit_on` a different stochastic scenario to the unit and the system does not break.

In what case would the `unit_flow` and `unit_on` use different stochastic scenarios? A dedicated stochastic scenario for a unit is set for `unit_on`. Isn't `unit_flow` always bonded with the stochastic scenario used by the `unit`? I'm actually not clear on how the stochastic scenario feature is implemented (e.g. how `constraint_operating_point_bound_indices` should look like?) Would extremely appreciate any further explanation!

nnhjy:
> This honestly looks great but it introduces an additional variable that is not needed if the efficiency curve (so to say) is monotonically decreasing. We could identify whether that's the case and then fallback to the previous implementation, but ideally I would like to do this as gracefully as possible. I might come up with some ideas very soon. In the meantime I think we can hold on this, if that's ok with you?

Just noticed that for the case where the efficiency curve is decreasing, there is a standalone approach implemented through [constraint_unit_pw_heat_rate](https://github.com/spine-tools/SpineOpt.jl/blob/cb26fa30e53cfb991ed4c5f2c281a167d969d9e3/src/constraints/constraint_unit_pw_heat_rate.jl).

src/constraints/constraint_operating_point_bounds.jl
nnhjy: Commit 13e6f70b7b23c3fe44f2027180a47e7b22558daf attempts to resolve the first concern on constraint indices thanks @manuelma for your help!

nnhjy: Improvement on this feature will continue in issue #631.

> One thing to keep in mind is how we would constrain e.g. different fuels in these settings. I guess the only way at this point would be to model this via 2 units with a designated user constraints. Alternatively we could also think of a constraint in the form of (draft example):
> 
> ```julia
> sum(unit_flow_op_active(u,n_i,d_i,op,s,t) for (u,n_i,d_i,op,s,t) in
> unit_flow_op_indices(unit=u,op=op,s=s,t=t)) <= 1 forall (u,op,s,t)
> ```
> 
> This could be an optional constraint to have a binary choice of which fuels and corresponding heat rates are used.

PR #541, opened 2023-02-17
Introduce explicit init_[variable_name] parameters

### LINKED ISSUE #455: Explicit initial_values
Currently we are using fix_* parameters to set the initial value for some variables. This can work but I feel it's not ideal and it's causing us problems:
 - It has to be set very precisely to have a value in the timeslice immediately preceding the start of the model horizon but must have a "none" value during the model horizon. If we change our model_start then it's not an intial value any more
 - If we want to use timeseries with ignore-year=true so that we don't have to uncessarily repeat the timeseries data, this causes problems too - because the initial value will repeat and we might not want that.

I feel the ideal solution is to simply have explicit initial values. So in addition to `fix_node_state`, for example, we will have, `initial_node_state`. And the same for every other variable that has a fix_* parameter. This way, SpineOpt will then internally always make sure that a fix_value is only set for the timeslice immediately preceding the model window start

manuelma:
Fixes #455

## Checklist before merging
- [ ] Unit tests pass


DillonJ:
It looks like you're fixing the variable in the first time slice rather than the last history timeslice of the first window... there may be arguments for do either... but my inclination is to fix the last history value

DillonJ:
Ah, ok - sorry, I read the code wrong then... I think that should be fine as that's what we are usually doing in practice. I'm thinking for investments for example - you want it to be able to invest at t0 but you want to force it to be zero in t-1

manuelma:
Ah, thanks for looking at the code @DillonJ !

I have validated it with a test so it's effectively fixing the last history time slice... Will merge as soon as test pass so we can move on. We can fix any indesired effects directly in master.

Issue #467, opened 2022-06-15
Return model object also on error


manuelma:

mihlema:
I think this also remotely links to a very old merge request #342, allowing users to print conflicts in case aa model is infeasible.
Probably something we should add in the future

DillonJ:
@manuelma I believe this is done, right? Can we close?

manuelma:
I don't think this is done? It was not straightforward as far as I remember...

clizbe:
@manuelma Is this now redundant with #794?

PR #543, opened 2023-02-21
Write intermediate 'as roll' results to binary files instead of the DB

### LINKED ISSUE #537: Performance of saving results degrades badly with rolling horizon
I have a model that rolls every 24 hours, meaning 365 windows to solve the full horizon.

When the model rolls, in the first windows, saving results (not writing, just saving) takes a couple of seconds. However it rises steadily each window to around 80 seconds in the final windows. That's roughly 4 hours of results saving for the full run.

We really need a new way to save results as the model rolls

manuelma:
Only write to DB at the very end.

This might improve performance of results writing with write-as-roll, since appending to an arrow file should be faster than writing to a Spine DB. Why? Because writing time series or any indexed value to a Spine DB implies reading the value, merging the new data, and then writing the value again. As more and more data is added to the time series, this reading/merging/writing becomes more and more expensive (more data to deal with at each time).

With the arrow technique, we don't need to do the reading/merging part, just the writing. But at the end we need to read from the arrow file, compile and write the time series to the DB as one large chunk, which should be somewhat expensive. We'll see - I think it's good to merge this PR for experimentation with proper real-world models.

Note that I'm using the apache arrow format as binary format to store our intermediate results but any format should do. Arrow seems nice because of the swapping technique they use to handle files larger than the RAM.

Fixes #537 #467 

## Checklist before merging
- [ ] Unit tests pass

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -4,6 +4,7 @@ authors = ["Spine Project consortium <spine_info@vtt.fi>"]
 version = "0.6.9"
 
 [deps]
+Arrow = "69666777-d1a9-59fb-9406-91d4454c9d45"
 Cbc = "9961bab8-2fa3-5c5a-9d89-47fab24efd76"
 Clp = "e2554f3b-3117-50c0-817c-e040a3ddf72d"
 DataStructures = "864edb3b-99cc-5e75-8d2d-829cb0a9cfe8"
--- a/Project.toml
+++ b/Project.toml
@@ -18,5 +18,5 @@ Requires = "ae029012-a4dd-5104-9daa-d747884805df"
 SpineInterface = "0cda1612-498a-11e9-3c92-77fa82595a4f"
 
 [compat]
-SpineInterface = "0.9.0"
+SpineInterface = "0.9.2"
 julia = "^1.2"
--- a/docs/src/library.md
+++ b/docs/src/library.md
@@ -19,6 +19,7 @@ Depth = 3
 
 ```@docs
 run_spineopt
+write_report_from_intermediate_results
 generate_forced_availability_factor
 ```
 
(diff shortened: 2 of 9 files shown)


DillonJ:
Does this mean that if the run is interrupted for any reason, we lose everything?

manuelma:
> @manuelma Does this mean that if the run is interrupted for any reason, we lose everything?

Why would that be the case @DillonJ ? I might be missing something...

DillonJ:
> > @manuelma Does this mean that if the run is interrupted for any reason, we lose everything?
> 
> Why would that be the case @DillonJ ? I might be missing something...

Maybe I'm issing something... but if we lose the run somehow, the results won't have been written to the db - that was the point of write_as_roll? Here it looks like you're only writing to the DB at the "very end". However I do see that you're saying this fixes the issue where we return the model object, even on error - so perhaps you are catching errors and then writing to the DB? However the point of write_as_roll is that if our PC crashes, or there's a power cut or whatever, we still have results in the DB up untill a certain point at least

manuelma:
> However the point of write_as_roll is that if our PC crashes, or there's a power cut or whatever, we still have results in the DB up untill a certain point at least

Thanks for the explanation! Absolutely, that's the main consideration here. I don't think we should do anything that contradicts that.

So I want to handle that in two ways: first, as you point out, we catch exceptions and write results from arrow files to the DB in the finally block (which runs even if an exception occurs - but might not run in case of power cut of course). This is already implemented as you observed.

Second, we don't delete the intermediate result files unless we're are sure we have written their contents to the DB. In this way the user can write them to the DB manually after the power cut (or any other eventuality). For this I am still to provide a API but should be straitghforward. My idea is to print the location of the arrow files to the user together with the specific function they need to call in order to write their contents to a DB.

The first is we never delete the arrow results file *unless* we have safely written their contents to the DB.

DillonJ:
> 

Great stuff @manuelma I think the payoff here is absolutely worth it. Excellent!

PR #566, opened 2023-03-08
Update constraint_units_on.jl

nnhjy:
Correct me if I misunderstood it.

`units_on` should align with `number_of_units` since `units_available` does it. See [constraint_units_available](https://github.com/spine-tools/SpineOpt.jl/blob/72800be43e4e39754f7cad5bce5e204e5277285a/src/constraints/constraint_units_available.jl),

Suppose `unitA` has 5 sub-units, all sharing a `unit_availability_factor` of 0.05. Then its `unit_available` would be 0.25. When its online variable is set to be an integer, no subunit is allowed to be online. 

## Checklist before merging
- [x] Unit tests pass


nnhjy:
@mihlema Many thanks for the feedback! The `unit_available` seem to play the role of `units_on` in the `constraint_unit_flow_capacity` as being the consequential coefficient for limiting the total aggregated available capacity. Why the  `unit_on` is used there then?

nnhjy:
Commit fa99c6f1602ecbff747325f7c7a8c793db4e5da6 is to solve conflicts in the constraints illustrated by the following instance:
```
split_ramps: -unit_flow[t1] + unit_flow[t2] - ramp_up_unit_flow[t2] + ramp_down_unit_flow[t2] = 0.0
minimum_operating_point : -56 units_on[t1] + unit_flow[t1] ≥ 0.0
unit_flow_capacity : -140 units_on[t2] + unit_flow[t2] ≤ 0.0
ramp_down : -70 units_on[t2] + 70 units_started_up[t2] + ramp_down_unit_flow[t2] ≤ 0.0
units_on[t1] = 7.0
units_on[t2] = 1.0
units_started_up[t2] ≥ 0.0
ramp_up_unit_flow[t2] ≥ 0.0
```
which ends up being
```
ramp_down_unit_flow[t2] - ramp_up_unit_flow[t2] ≥ 252
ramp_down_unit_flow[t2] ≤ 70
```
The problem is that the decrease of online units is not considered in the constraint on `ramp_down_unit_flow` (i.e. the "ramp_down").

nnhjy:
Commit 7a9ea0d2caac2c34d82dc695d5ccbee649e0be5a is to solve conflicts in the constraints illustrated by the following instance:
```
unit_state_transition: -units_on[t1] + units_on[t2] - units_started_up[t2] + units_shut_down[t2] = 0.0
split_ramps: -unit_flow[t1] + unit_flow[t2] - ramp_up_unit_flow[t2] + ramp_down_unit_flow[t2] = 0.0
minimum_operating_point : -56 units_on[t2] + unit_flow[t2] ≥ 0.0
unit_flow_capacity : -140 units_on[t1] + unit_flow[t1] ≤ 0.0
ramp_up : -70 units_on[t2] + 70 units_started_up[t2] + ramp_up_unit_flow[t2] ≤ 0.0
units_on[t1] = 1.0
units_on[t2] = 13.0
units_shut_down[t2] ≥ 0.0
ramp_down_unit_flow[t2] ≥ 0.0
```
This can be reformulated into the following:
```
unit_state_transition: units_started_up[t2] - units_shut_down[t2] = 12.0
split_ramps: -unit_flow[t1] + unit_flow[t2] - ramp_up_unit_flow[t2] + ramp_down_unit_flow[t2] = 0.0
minimum_operating_point : unit_flow[t2] ≥ 56 * 13.0
unit_flow_capacity : unit_flow[t1] ≤ 140
ramp_up : 70 units_started_up[t2] + ramp_up_unit_flow[t2] ≤ 70 * 13.0
```
with `units_started_up[t2]` = 12.0, we obtain
```
ramp_up_unit_flow[t2] - ramp_down_unit_flow[t2] ≥ 56 * 13 - 140 = 588
ramp_up_unit_flow[t2] ≤ 70
```
meaning that `ramp_down_unit_flow[t2]` must be negative.

The problem originates from constraint `constraint_ramp_up` omitting the `unit_flow` increase attributed to started-up units.

mihlema:
Hi @nnhjy 

Sorry I didn't get pinged for the ramping issue you encountered. IMO, the issue does not arise from the formulation itself but from the (somewhat) tricky usage (apologies). When the unit flow is split into multiple reserves, it is not only split into ramp_up and ramp_down but also into start_up and shut_down ramps. The problem that you described should therefore not be encountered. The description of the usage is [here](https://spine-tools.github.io/SpineOpt.jl/latest/advanced_concepts/ramping_and_reserves/). Not sure how clearly defined it is here, but you will always need to also define max_shutdown/startup_ramp and min_shutdown/startup_ramp as 1 and 0, respectively. (or lower/higher values if you are constraining your startup/shutdown ramps).

I then also came across this pull request and noticed that you made some changes with regards to the ramping constraints. I don't fully understand all of these changes but I'm probably just missing something here, I hope you can help me here, I'm going to compare the new and the old version here (simplified indices, removed non-spinning reserves):

Original formulation:
```
@constraint(
            m,
                ramp_down_unit_flow[u, n, d, s, t]
            <=
            + sum(
                (
                    + units_on[u, s, t1]
                    - units_started_up[u, s, t1]
                    )
                )
                * duration(t_on)
                * ramp_down_limit[(unit=u, node=ng, direction=d, stochastic_scenario=s, analysis_time=t0, t=t)]
                * unit_conv_cap_to_flow[(unit=u, node=ng, direction=d, stochastic_scenario=s, analysis_time=t0, t=t)]
                * unit_capacity[(unit=u, node=ng, direction=d, stochastic_scenario=s, analysis_time=t0, t=t)]
        )
```
Modified version:
``` 
@constraint(
            m,
                ramp_down_unit_flow[u, n, d, s, t]
            <=
                        sum(+ units_on[u, s, t1] - units_started_up[u, s, t1]
                    )
                    * ramp_down_limit[(unit=u, node=ng, direction=d, stochastic_scenario=s, analysis_time=t0, t=t)]
                    + units_shut_down[u, s, t1]
                )
                * duration(t_on)
                * unit_conv_cap_to_flow[(unit=u, node=ng, direction=d, stochastic_scenario=s, analysis_time=t0, t=t)]
                * unit_capacity[(unit=u, node=ng, direction=d, stochastic_scenario=s, analysis_time=t0, t=t)]
        )
``` 

For me the right hand-side of this constraint is not logical, but maybe I am overlooking something here. How this constraint was supposed to work is that on the left-hand side you'd have the possible spinning ramp (i.e., not start-up nor shutdown ramp), constrained by the number of spinning units (i.e., units that are not currently starting up or are not shutting down) multiplied by the ramping down limit (again, spinning down limit, not shut_down_ramp). The **` units_shut_down`**  should IMO not be added here (this is defined in another constraint explicit for shut_down ramps: [constraint_max_shut_down_ramp.jl](https://github.com/spine-tools/SpineOpt.jl/blob/master/src/constraints/constraint_max_shut_down_ramp.jl) and [constraint_min_shut_down_ramp.jl](https://github.com/spine-tools/SpineOpt.jl/blob/master/src/constraints/constraint_min_shut_down_ramp.jl).

Unless this has been already thoroughly reviewed, I would strongly **recommend reverting this merge commit**, and first ensure that we are not breaking anything/we're sure we're changing it correctly. Maybe I also missed a discussion on this and maybe it's already been thoroughly reviewed. @manuelma I think you were possibly reviewing, not sure?

nnhjy:
Thank you @mihlema for reviewing this and pointing out the issue! You are right. The added `units_shut_down` would duplicate in a naive manner the two constraints for shut down ramp if the parameters `max/min_shutdown_ramp` are defined. In that case, the formulations become incorrect.

Nevertheless, instead of going back to the original version, perhaps we could/should still allow the model to work when the parameters `max/min_shutdown_ramp` are not defined. They are by default `None` to avoid generating unnecessary constraints as I see. But since the variables `units_start_up` and `units_shut_down` are always used, the constraints for "start up" and "shut down" ramping are necessary to model ramping. And so all 6 limit parameters for both layers of "ramp" and "start/shut ramp" are required (as our standard [instruction](https://spine-tools.github.io/SpineOpt.jl/latest/advanced_concepts/ramping_and_reserves/#Ramping-and-Reserves) tells, or maybe it just needs 3 to do it for single direction?) to make the model work correctly. Users who are lazy/less professional/at an early stage (like me for example) would easily get lost (and I am luckily saved by your nice explanation : ) ).

I've just come up with two initial thoughts:
1. to make [constraint_max_shut_down_ramp.jl](https://github.com/spine-tools/SpineOpt.jl/blob/master/src/constraints/constraint_max_shut_down_ramp.jl) and [constraint_min_shut_down_ramp.jl](https://github.com/spine-tools/SpineOpt.jl/blob/master/src/constraints/constraint_min_shut_down_ramp.jl) (likewise for the start up ramp constraints) activated by default along with the ramping constraints, i.e. when the ramping constraints are implemented, the corresponding "max limit" parameters for start up/shut down ramp should be set to 1 if not defined.
2. to make the added `units_shut_down` conditional to whether the max limit parameter for shut down ramp is defined, e.g.:
```julia
@constraint(
 #           m,
 #               ramp_down_unit_flow[u, n, d, s, t]
 #           <=
 #             (
 #                     sum(+ units_on[u, s, t1] - units_started_up[u, s, t1])
 #                   * ramp_down_limit[(unit=u, node=ng, direction=d, stochastic_scenario=s, analysis_time=t0, t=t)]
                    + (max_shutdown_ramp==nothing ? units_shut_down[u, s, t1] : nothing)
 #               )
 #              * duration(t_on)
 #              * unit_conv_cap_to_flow[(unit=u, node=ng, direction=d, stochastic_scenario=s, analysis_time=t0, t=t)]
 #              * unit_capacity[(unit=u, node=ng, direction=d, stochastic_scenario=s, analysis_time=t0, t=t)]
        )
```
, which is equivalent to assuming `max_shut_down_ramp` to be 1 when it is not defined. This approach requires fewer constraints.

Both ways could relieve some tedious work for users who want to model ramps in different detail while staying with our awesome ramp formulations.

mihlema:
Thanks @nnhjy, I agree with you that the current work around is not very user -friendly! I had intended to include default ramp rates, as you are suggesting, to make life easier, but shied away from it to reduce number of constraints. Your option one makes sense to me but I agree that it'd be overly expensive in terms of number of constraints.

Regarding option two I think that could work, but feels somewhat dangerous in terms of defining the same thing in two places. - However, would reduce the number of variables and constraints, so it might be okay.

I think the following solution might also be robust, and the same time doesn't flood the model too much with unnecesary constraints:
- add method parameter "use_ramps" on unit level (or unit__to_node/from_node level if we want to be generic)
- add default for ` max_shutdown_startup_limits`  and ` ramp_up/down`  and ` min_shutdown_startup_limits`   (1 and 0, respectively)
Only if ` use_ramps` is set to ` true` create (all) ramp variables and constraints. This would mean we'd have some extra variables/constraints in case we are not using startup ramps, but it should be pretty failsafe. WDYT?

To use ramps the user would only need to set "use_ramps" to true (we could even think of having this as a global parameter optionally) and adapt the ramping parameters of interest.

nnhjy:
Nice idea too! I agree this method feels more standard approach. But as there are going to be some fundamental changes on the ramping formulations, perhaps we could take the simplest fix to avoid rewriting the ramping sections of the documentation.

I also discussed this with @manuelma. Maybe he could provide some thought.

mihlema:
I think indeed we can wait what comes out of the discussion. Ideally we would at least flag this in the code, that this functionality is broken through this merge request - I am not a big fan of letting broken things lurke in the code however.
For next PR, ideally if it's possible, could re-request for another review if you change something after review? 
But I know sometimes this process might take a while unfortunately.

nnhjy:
@mihlema Good idea and thanks for the suggestion! Will do it that way. I wouldn't like to break the functionality, either. 

> For next PR, ideally if it's possible, could re-request for another review if you change something after review?
> I am not a big fan of letting broken things lurke in the code however.

I will give it a quick fix for the current implementation.

PR #597, opened 2023-04-08
Use new SpineInterface API in preprocessing

manuelma:
## Checklist before merging
- [ ] Release notes have been updated
- [ ] Code has been formatted nicely
- [ ] Unit tests pass


DillonJ:
This sounds big - have we done a little systems testing?

manuelma:
It's actually very small, even if a lot of lines have changed.

But good point about the system testing. We need to set something up soon, if only we had more resources - maybe within Mopo?

Issue #647, opened 2023-05-21
Keep documentation for current implementations


nnhjy:
Concerning the documentation of some formulations and functionalities that have recently been removed in commits 6ccbb14, 46eda5e, 17a3d14, 3f00373, I suggest we restore and keep them for now for the reasons that 
1) the rollout of the new implementation still needs some time 
2) the current implementation as such still exists and works. 

The users would appreciate those clues on what is going on at the moment. And it would not be too late to replace them with the new ones with the PR for new implementation. WDYT?

tarskul:
Confusion was exactly the reason why they were removed in the first place. It is not necessarily that the formulation was incorrect but the formulation combined different conditions into one formulation making people who are not used to the code and the formulation believe that the formulation was incorrect (e.g. upward and downward spinning reserves in the same equation even though they never show up in the same equation during runtime as only parts of the equation are activated at the same time). So, it is probably best that at the very least they are clarified before returning them (or we wait for the new changes which should not take to long (?)).

DillonJ:
I think the issue here is that there is the formulation and there is the implementation. It appears that in this instance the formulation was correct and the implementation was correct. The issue was that it was a little difficult in this instance to make the connection between the two. I think this is a developer issue - I feel that a normal user will not necessarily need to understand the implementation of the fomulation. Additionally, there are unit tests that verify that constraints generated are consistent with the formulation.

The question is, do we need to get into implementation details in the documentation and if we do, will that confuse users even more? Perhaps some additional text in the docstring to explain what is happening?

manuelma:
We never want to document implementation details. We want to be able to change the implementation at will, without worrying about updating the documentation at the same time.

jkiviluo:
@manuelma That's wise. So, anything that explains implementation should be in docstrings - and is meant for model coders.

manuelma:
@jkiviluo actually, *docstrings* are meant to documentation and documentation alone. It's a string that you put below or above a function definition where you describe the usage (purpose, parameters, return value, etc). To me, ideally we would describe implementation as *comments* (not the same as docstrings) inline the code itself.

tarskul:
Unless we make different regions in the docstring as suggested as one of the possible design decisions in issue #649 (e.g. #region developernote). That way we can still later decide programmatically whether we want the implementation details in the documentation or not.

jkiviluo:
[...]

PR #648, opened 2023-05-21
Fix broken ramp up and down formulations

nnhjy:
A workaround fix w.r.t. [proposal 2](https://github.com/spine-tools/SpineOpt.jl/pull/566#issuecomment-1533060728) on the [issue](https://github.com/spine-tools/SpineOpt.jl/pull/566#issuecomment-1532304247) discussed under PR #566. The aim is to fix the broken functionality while minimising changes and users' application efforts as new implementation is pending. A more standardized option can be found [here](https://github.com/spine-tools/SpineOpt.jl/pull/566#issuecomment-1533145163)

## Checklist before merging
- [x] Unit tests pass


nnhjy:
The documentation of the formulations has currently been removed (#647 ). I will update it as soon as it gets restored.

mihlema:
Hi @nnhjy, thanks for taking care of this. I quickly looked through this and the fix seems to make sense to me. Pending the documentation status, we could merge this.
To summarize what this merge request will enable: If `max_startup/shutdown_ramp` parameters are defined, they will be used as an upperbound for ramping during start/shutdown of a unit. If they are not defined, we assume that the unit can ramp up/down full capacity.

test/constraints/constraint_unit.jl
```diff
@@ -1428,6 +1428,31 @@ function test_constraint_ramp_up()
             observed_con = constraint_object(constraint[con_key...])
             @test _is_constraint_equal(observed_con, expected_con)
         end
+
+        max_startup_ramp = 0.4
```
mihlema, line 1432: Is max_startup_ramp actually used in this test? Seems like the constraint function actually calls ramp_up_limit. Could we have tests that cover the two cases 1) max_startup_ramp defined/not defined?

mihlema: Actually, I guess you are testing the "is defined" case here, and probably the not-defined case has been added beforehand, so should be good

nnhjy: Thanks for the review! Exactly, this block tests the case where `max_startup_ramp` is defined. And the previous block tests the case where it is not defined with the default value `None`.

nnhjy:
Hi @mihlema the documentation has got returned and renewed accordingly. I hope the formulations for this change is correctly written.

PR #666, opened 2023-06-12
Issue 656 simpler benders

### LINKED ISSUE #656: Simplify the benders data structure
So currently, to use benders decomposition one needs to specify a new model object to represent the benders master problem, and give it `model_type_benders` as type. But in addition, one also needs to create all the model__temporal_block, model__investment_temporal_block and so on, for the 'benders' model object. Is there a case where these would difer from the ones for the 'standard' model object? My impression is that the opposite is true: one needs to replicate the exact same relationships in the 'standard' and the 'benders' object, otherwise funny things may happen.

If my impression is true, I would like to consider triggering the benders decomposition differently, by a parameter in the 'standard' model object itself. @DillonJ what do you think?

manuelma:
Active benders by simply selecting model_type spineopt_benders.

Fixes #656

## Checklist before merging
- [x] Unit tests pass


manuelma:
We need units_on to be a variable for benders to work, so we need to check that variable_type_none is *not* chosen for any unit with candidate_units > 0.

PR #662, opened 2023-06-08
618 add how to section to documentation

tarskul:
Added new chapters:
+ How To
+ Implementation detail

Also there are some minor details that needed to be changed to better accommodate the tutorial chapter.

This pull request might be unnecessary as the pull request draganddrop contains the same changes and the draganddrop feature. They are only made separate in case the new chapters are accepted but the draganddrop feature is not.

Fixes #618

## Checklist before merging
- [x] Documentation is up-to-date
- [x] Unit tests have been added/updated accordingly
- [x] Code has been formatted according to SpineOpt's style
- [x] Unit tests pass

PR #693, opened 2023-07-05
Draganddrop

### LINKED ISSUE #618: Add How to Section to Documentation
We have discussed many times that the current documentation is good as a reference when you know what you are looking for - but it it's not so good at explaining how to do specific high-level things that might involve multiple elements.

For example, how to change the solver (https://github.com/spine-tools/SpineOpt.jl/issues/616) or how to define an efficiency (https://github.com/spine-tools/SpineOpt.jl/discussions/617).

I was thinking that we could add a "How To" section to the documentation and add articles which in many cases the content may exist already in issues or discussions. Ideally, it would be pretty easy to add these articles so they can be easily added to over time. For example, if we could have a how to folder and just drop in MD files of the relevant how_to article and the contents/index could be generated automatically based on the contents of the How To folder.

The two issues/discussions above would be a great start.

I have added an issue label "Add how to article to docs" so we can tag issues that may have useful content for a how-to article.

tarskul:
Added new chapters:
+ How To
+ Implementation detail

Also there are some minor details that needed to be changed to better accommodate the tutorial chapter.

For select chapters there is now a drag-and-drop feature. For those chapters you can simply add your markdown file to the folder of the chapter and it will be automatically added to the documentation. To allow both manually composed chapters and automatically generated chapter, the functionality is only activated for empty chapters (of the structure "chapter name" => nothing).

This branch builds further on another branch related to #618 so keep that in mind when adding these two pull requests.

Fixes #618 

## Checklist before merging
- [x] Unit tests pass


manuelma:
Nice one @tarskul !

One thing I would recommend is to slightly fix the style to match SpineOpt's. In your code, I have noticed that you don't put spaces around equal signs (=) or after commas (,).

So for instance, please consider replacing `array=[1,2,3]` with `array = [1, 2, 3]` - note the spaces around equal and after commas. Similary, `function x(a,b,c)` can be written as `function x(a, b, c)`. This will make code much more consistent and eventually easier to read.

tarskul:
Right, I did notice this style in SpineOpt and I told myself to try and follow it but it seems that the force of habit is quite strong. I'll make some changes.

manuelma:
> Right, I did notice this style in SpineOpt and I told myself to try and follow it but it seems that the force of habit is quite strong. I'll make some changes.

thank you very much!

manuelma:
To make things slightly more complex, we don't put spaces around equal signs in keyword arguments, so we write `function_name(args..., kw=something)`

Why make it simply if you can make it complex right? I believe this convention comes from python pep8...

Tasqu (changes requested):
As far as I can tell, besides the missing figures like in #662 this is good to go for merge.

Tasqu (approved):
There seem to be some old figure referencing and cross-referencing issues, but as far as I can tell they are also plaguing the `master` branch. We should probably deal with them eventually, but not really a problem for this branch. Good to go, as far as I'm concerned.

PR #707, opened 2023-07-14
Fix duals calculation for Benders

manuelma:
Previously, we were computing the marginal value of the unit investments as the reduced cost of the units_on variable. This is a bit non-intuitive - it should be the reduced cost of units_invested_available.

But units_invested_available didn't work for us reliably, and I think I just found out why:
when computing the duals, if units_on was discrete in the original problem, we would fix its value in the relaxed problem. That means additional investments wouldn't be usable (because the units_on was already fixed) resulting in reduced cost of units_invested_available being zero.

This commit rationalizes the whole thing. The marginal value of the unit investments is obtained from the reduced cost of units_invested_available, as it should be. To avoid it being zero, we don't fix discrete variables to calculate duals - we just relax them.

The above only applies to computing duals for creating Benders cuts; for reporting purposes we use the regular, tested method that relaxes integrality and then fixes the values.

Note than when using CPLEX, we can't use the internal routine to solve the relaxed problem to calculate duals for Benders, because it fixes the discrete variables and we just can't control it. CPLEX wasn't working properly for Benders anyway - with this commit, it works.

## Checklist before merging
- [x] Unit tests pass

### DIFF
--- a/src/constraints/constraint_mp_any_invested_cuts.jl
+++ b/src/constraints/constraint_mp_any_invested_cuts.jl
@@ -45,7 +45,8 @@ function add_constraint_mp_any_invested_cuts!(m::Model)
                 )
                 * sum(
                     Iterators.filter(
-                        !isnan, units_on_mv(unit=u, stochastic_scenario=s, t=t) for t in t_overlaps_t(m; t=t)
+                        !isnan,
+                        units_invested_available_mv(unit=u, stochastic_scenario=s, t=t) for t in t_overlaps_t(m; t=t)
                     );
                     init=0
                 )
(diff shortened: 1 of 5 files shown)


DillonJ:
This sounds very promising @manuelma. A couple of questions/thoughts...
 - Is this only relevant for Benders? I.e. in a "standard" model, we fix the integer variables to obtain duals?
 - Have we tested this from before the change - do we get the same solutions?

manuelma:
Thank you @DillonJ I can answer the first question already, yes, this PR only affects the duals calculation for the purpose of creating benders cuts. The standard model calculates the duals the usual with way, by both relaxing and fixing the discrete variables...

The second question is the important one before we are able to merge this. Will make sure it is the case (that we get the same solutions) before merging.

manuelma:
This works, results are exactly the same.

Issue #654, opened 2023-05-31
Improve formulations and implementations for ramping and reserves


nnhjy:
We would like to make it as "correct, fast, and flexible to use" as possible ...

Specific actions:

### Documentation 
- [x] Add a tutorial on ramping and reserve use in SpineOpt
- [ ] Create a table in the documentation that shows which constraints appear in SpineOpt when we have specific parameters.
- [x] Add the constraint "unit flow capacity with ramps" to the documentation and rename it to avoid misleading the users (the name suggests ramping constraint, but is creating an upper bound for the capacity)

### Constraint unit flow capacity with ramps
- [x] Include the unit committed for the non-spinning downward reserve in this constraint according to the proposed formulation.
  - [x] Change the constraint
  - [x] Modify/add tests
  - [x] Change the documentation
- [x] Double-check the activation of this constraint since, in a simple example appears the “constraint_unit_flow_capacity” (basic version), instead of this detailed version.

### Constraint on minimum operating point
WIP: branch 654-constraint-minimum-operating-point
- [x] It is necessary to subtract the downward reserves (including the non-spinning reserve) in LHS and the unit committed to the spinning reserve in the RHS
  - [x] Change the constraint
  - [x] Modify/add tests
  - [x] Change the documentation
  
### Ramp up, ramp down, and split ramps constraints

- [x] The RHS is different in SpineOpt for both ramping constraints from the proposed formulation
- [x] Since the reserves are treated as nodes, unnecessary constraints appear in the model. So, we need to clean the code to only create the necessary constraints. For example, in the ramp-up constraint appears:
Flow ramp + upwards reserve <= RHS (correct)
Flow ramp - downwards reserve <= RHS (unnecessary)
- [x] Double-check possible changes in other existing constraints in SpineOpt related to ramps, such as max_start_up_ramp, max_shut_down_ramp, min_start_up_ramp, and min_shut_down_ramp.

Note: Issue #603 looks relevant.

DillonJ:
Some thoughts on this.

Some high level comments that I think are important:

 - I think we should support reserves that are not linked to ramping - this is more computationally efficient. For example, just to constraints: 1. Reserve < capacity - flow (headroom constraint) and 2. Reserve < capability (where capability is the maxium capability over the reserve time frame that might be related to ramping but also other considerations but there is no explicit link with ramping (computationally efficient).
[...]

PR #742, opened 2023-08-28
Change constraint min operating point, and its test and doc

gnawin:
The constraint minimum operating point and its test and documentation on the math formulation have been changed.

It closes part of #654. 

## Checklist before merging
- [x] Unit tests pass


gnawin:
Hi @mihlema @DillonJ, I think this change is interesting for you, so I requested you both as reviewers. Would you please have a look? Feel free to propose others in place/additional as well. Thanks!

manuelma:
I propose I review this.

@gnawin would it be possible to describe the change in words? What changed in the constraint and why?

manuelma:
I think I get the idea. 
Before the constraint was, roughly:

```
unit_flow >= min_op_point * units_on
```

And now it is, again, roughly:
```
normal_unit_flow - downward_reserve_unit_flow >= min_op_point * (units_on - units_hold_for_downward_reserve_provision)
```

Which makes sense to me. So we don't count the downward reserve provision for the purposes of imposing the unit's minimum operating point.

Is that about right @gnawin ?

gnawin:
> I think I get the idea. Before the constraint was, roughly:
> 
> ```
> unit_flow >= min_op_point * capacity * units_on
> ```
> 
> And now it is, again, roughly:
> 
> ```
> normal_unit_flow - downward_reserve_unit_flow >= min_op_point * capacity * (units_on - units_held_for_downward_reserve_provision)
> ```
> 
> Which makes sense to me. So we don't count the downward reserve provision for the purposes of imposing the unit's minimum operating point.
> 
> Is that about right @gnawin ?

Yes, that's basically the idea. See more information below.

The new formulation is given in the documentation within this PR. 

Major change 1:
The indices of the constraint remain the same, but a few terms on both the LHS and the RHS have been added. 
On the LHS, besides the flows for the non-reserves, the flows for the reserves (downward spinning and non-spinning) has been substracted. 
On the RHS, the unit committed to the non-spinning reserve has been substracted for the units on.

Major change 2:
On the LHS, `duration(t)` was outside the sume of `unit_flow`. This was wrong because in case of different time resolution for `unit_flow`, `duration(t)` has to be inside the sum as well. 

Question:  
However, the t_indices may need to be changed, see #763.

The unit tests has been modified, and written in a way that reflects the major change 2.

DillonJ:
Questions @manuelma if I don't have any reserves defined, does this add new variables to the problem that aren't needed?

manuelma:
> Questions @manuelma if I don't have any reserves defined, does this add new variables to the problem that aren't needed?

Good question @DillonJ - no, it doesn't. This PR is very slim, it doesn't change anything related to variable generation - just adds a couple of extra terms in one constraint.

manuelma:
Nice work @gnawin !

PR #753, opened 2023-09-25
WIP Resolve JuMP `Parameter` and `parameter_value` conflicts

### LINKED ISSUE #745: JuMP compatibility conflict: Parameter used both by JuMP and SpineInterface
The latest version of JuMP exports Parameter which conflicts with SpineInterface.

To resolve that issue we can either import each function of a module separately (excluding the ones we don't need) or we can follow the pythonic approach and import the module instead of using the module.

Tasqu:
# THIS PULL REQUEST IS STILL A WORK IN PROGRESS! DO NOT MERGE!

This pull request addresses the `Parameter` and `parameter_value` name conflict with SpineInterface in JuMP v1.14 and higher. Resolving the name conflict wasn't terribly difficult, but there are a few odd errors remaining.

Fixes #745 (eventually?)

## Checklist before merging
- [x] Unit tests pass

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -19,4 +19,3 @@ SpineInterface = "0cda1612-498a-11e9-3c92-77fa82595a4f"
 [compat]
 SpineInterface = "0.11.6"
 julia = "^1.2"
-JuMP = "< 1.14"
--- a/src/SpineOpt.jl
+++ b/src/SpineOpt.jl
@@ -33,6 +33,9 @@ import DataStructures: OrderedDict
 import Dates: CompoundPeriod
 import LinearAlgebra: BLAS.gemm, LAPACK.getri!, LAPACK.getrf!
 
+# Resolve JuMP and SpineInterface `Parameter` and `parameter_value` conflicts.
+import SpineInterface: Parameter, parameter_value
+
 export run_spineopt
 export rerun_spineopt
 export run_spineopt_kernel!
--- a/src/constraints/constraint_unit_flow_op_bounds.jl
+++ b/src/constraints/constraint_unit_flow_op_bounds.jl
@@ -32,16 +32,16 @@ function add_constraint_unit_flow_op_bounds!(m::Model)
             + unit_flow_op[u, n, d, op, s, t]
             <=
             (
+                ordered_unit_flow_op(unit = u, node=n, direction=d, _default=false) ? 
+                unit_flow_op_active[u, n, d, op, s, t] : units_on[u, s, t]
+            )
+            * (
                 + operating_points[(unit=u, node=n, direction=d, stochastic_scenario=s, analysis_time=t0, i=op)] 
                 - (
                     (op > 1) ? operating_points[
                         (unit=u, node=n, direction=d, stochastic_scenario=s, analysis_time=t0, i=op - 1)
                     ] : 0
                 )
-            )
-            * (
-                ordered_unit_flow_op(unit = u, node=n, direction=d, _default=false) ? 
-                unit_flow_op_active[u, n, d, op, s, t] : units_on[u, s, t]
             )
             * unit_availability_factor[(unit=u, stochastic_scenario=s, analysis_time=t0, t=t)]
             * unit_capacity[(unit=u, node=n, direction=d, stochastic_scenario=s, analysis_time=t0, t=t)]
--- a/src/constraints/constraint_units_available.jl
+++ b/src/constraints/constraint_units_available.jl
@@ -32,15 +32,15 @@ function add_constraint_units_available!(m::Model)
                 units_available[u, s, t] for (u, s, t) in units_on_indices(m; unit=u, stochastic_scenario=s, t=t);
                 init=0,
             )
-            <=
-            + number_of_units[(unit=u, stochastic_scenario=s, analysis_time=t0, t=t)] 
-            + expr_sum(
+            - expr_sum(
                 units_invested_available[u, s, t1]
                 for (u, s, t1) in units_invested_available_indices(
                     m; unit=u, stochastic_scenario=s, t=t_overlaps_t(m; t=t)
                 );
                 init=0,
             )
+            <=
+            number_of_units[(unit=u, stochastic_scenario=s, analysis_time=t0, t=t)] 
         )
         for (u, s, t) in constraint_units_available_indices(m)
     )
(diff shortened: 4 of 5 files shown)


Tasqu:
I've been looking into this for a while now, and it seems whatever is causing our mystery error seems to be caused by the `add_constraint_unit_flow_op_bounds!` function. Furthermore, it indeed seems to be related to multiplication, as changing `* -> +` avoids the errors, although fails the unit tests as the constraint is different. The problematic multiplication seems to occur in this part:
```julia
          (
              + operating_points[(unit=u, node=n, direction=d, stochastic_scenario=s, analysis_time=t0, i=op)] 
              - (
                  (op > 1) ? operating_points[
                      (unit=u, node=n, direction=d, stochastic_scenario=s, analysis_time=t0, i=op - 1)
                  ] : 0
              )
          )
          * (
              ordered_unit_flow_op(unit = u, node=n, direction=d, _default=false) ? 
              unit_flow_op_active[u, n, d, op, s, t] : units_on[u, s, t]
          )
```

Tasqu:
I have zero idea why, but rearranging the two terms in the problematic constraints seems to solve the issue:
```julia
            (
                ordered_unit_flow_op(unit = u, node=n, direction=d, _default=false) ? 
                unit_flow_op_active[u, n, d, op, s, t] : units_on[u, s, t]
            )
            * (
                + operating_points[(unit=u, node=n, direction=d, stochastic_scenario=s, analysis_time=t0, i=op)] 
                - (
                    (op > 1) ? operating_points[
                        (unit=u, node=n, direction=d, stochastic_scenario=s, analysis_time=t0, i=op - 1)
                    ] : 0
                )
            )
```
works

Tasqu:
So I've got a lead on to what's causing our one remaining issue _(assuming I haven't accidentally created new ones)_: The "problem" is the addition of terms on the right-hand side of the `add_constraint_units_available!`
```julia
function add_constraint_units_available!(m::Model)
    @fetch units_available, units_invested_available = m.ext[:spineopt].variables
    t0 = _analysis_time(m)
    m.ext[:spineopt].constraints[:units_available] = Dict(
        (unit=u, stochastic_scenario=s, t=t) => @constraint(
            m,
            + expr_sum(
                units_available[u, s, t] for (u, s, t) in units_on_indices(m; unit=u, stochastic_scenario=s, t=t);
                init=0,
            )
            <=
            + number_of_units[(unit=u, stochastic_scenario=s, analysis_time=t0, t=t)] 
            + expr_sum(
                units_invested_available[u, s, t1]
                for (u, s, t1) in units_invested_available_indices(
                    m; unit=u, stochastic_scenario=s, t=t_overlaps_t(m; t=t)
                );
                init=0,
            )
        )
        for (u, s, t) in constraint_units_available_indices(m)
    )
end
```
It seems that for whatever reason, the `+ expr_sum(...)` does something to the expression that causes the constraint to be evaluated as `0 <= 0`, even though it used to be evaluated as `nothing`. The lower `expr_sum` actually does nothing in the unit test, and removing it allows for the test to pass.

Seems like the issue is somehow related to operations between `Call` and `AffExpr`, but I'm not sure where exactly.

Tasqu:
... And again, this unit test can be "fixed" by rearranging the terms:
```julia
function add_constraint_units_available!(m::Model)
    @fetch units_available, units_invested_available = m.ext[:spineopt].variables
    t0 = _analysis_time(m)
    m.ext[:spineopt].constraints[:units_available] = Dict(
        (unit=u, stochastic_scenario=s, t=t) => @constraint(
            m,
            + expr_sum(
                units_available[u, s, t] for (u, s, t) in units_on_indices(m; unit=u, stochastic_scenario=s, t=t);
                init=0,
            )
            - expr_sum(
                units_invested_available[u, s, t1]
                for (u, s, t1) in units_invested_available_indices(
                    m; unit=u, stochastic_scenario=s, t=t_overlaps_t(m; t=t)
                );
                init=0,
            )
            <=
            number_of_units[(unit=u, stochastic_scenario=s, analysis_time=t0, t=t)] 
        )
        for (u, s, t) in constraint_units_available_indices(m)
    )
end
```
Even though these changes technically "fix" things, I don't think these changes address the underlying issues properly. It's far to easy to run into weird bugs and generate useless constraints if the order of the terms affects the outcome.

As far as I can tell, this is related to how JuMP `@constraint` macro handles `Call` and `AffExpr`, so maybe these can be tweaked via `src/update_model.jl`

DillonJ:
@Tasqu nice work, indeed, it's rather unsatifying "fixing" something when you've only a vague idea why it fixed it! 

@manuelma any ideas here?

Tasqu:
Seems like the `build_constraint` extension didn't cover enough cases, so the `test_unit_online_variable_type_none` was failing due to generating a useless constraint instead of returning `nothing`. By adding the following methods:
```julia
function JuMP.build_constraint(_error::Function, expr::AffExpr, set::_CallSet)
    build_constraint(_error, Call(1) * expr, set)
end
function JuMP.build_constraint(_error::Function, expr::AffExpr, lb::Real, ub::Call)
    build_constraint(_error, Call(1) * expr, Call(lb), ub)
end
function JuMP.build_constraint(_error::Function, expr::AffExpr, lb::Call, ub::Real)
    build_constraint(_error, Call(1) * expr, lb, Call(ub))
end
function JuMP.build_constraint(_error::Function, expr::AffExpr, lb::Call, ub::Call)
    build_constraint(_error, Call(1) * expr, lb, ub)
end
```
we can cover this case at least. Seems like the strict typing of
```julia
function JuMP.add_constraint(
    model::Model,
    con::ScalarConstraint{GenericAffExpr{Call,VariableRef},S},
    name::String="",
) where {S<:_CallSet}
    iszero(con.func) && return nothing
    con_ref = Ref{ConstraintRef}()
    realized_func = realize(con.func, con_ref)
    realized_set = realize(con.set, con_ref)
    realized_constraint = ScalarConstraint(realized_func, realized_set)
    con_ref[] = add_constraint(model, realized_constraint, name)
end
```
was causing an issue in this case, where the generated useless `0 <= 0` `con` was not of type `ScalarConstraint{GenericAffExpr{Call,VariableRef},_CallSet}`, and thus didn't pass through this method where useless constraints are removed by `iszero(con.func) && return nothing`. Instead, some standard `JuMP.build_constraint` method was used, and thus the constraint ended up being generated.

Unfortunately, I still have zero idea about the `operate!(*, ::GenericAffExpr{Call, VariableRef}, ::Call) is not implemented yet.` error.

PR #764, opened 2023-10-02
649 documentation from docstring

### LINKED ISSUE #649: Build documentation from docstrings (e.g. constraints)
To streamline the development of new features and the documentation thereof, it could be interesting to build the documentation from the docstrings instead of making separate documentation files. This approach could also help to make sure that the documentation unit test does not fail (issue [#626](https://github.com/spine-tools/SpineOpt.jl/issues/626)).

I've already started an experiment with the constraints to look for possibilities for implementing this approach. From that experiment a few implementation considerations became clear.

The first is that the constraints are written in Latex which julia docstrings do not like due to the many escaped characters. One way around this is to escape the characters correctly. However that might be confusing when you are used to writing Latex regularly. There is another way which allows to write Latex within a special docstring provided by the Documenter package. However, that would require to add Documenter as a dependency of the SpineOpt tool.

Regardless of the Latex characters, there are also two ways to build the documentation from the docstring. The first is quite straightforward. Documenter allows to call the docstring documentation directly in your documentation file. Although that should be rather straightforward, the result might not look too great. That is unless we deviate from the typical structure where the function and the corresponding parameters are highlighted. Another approach is to add special fields to the docstring and search for those special fields. In the case of the constraints that means that there will at least be two fields: one for the formulation and one for a short description thereof. I propose to mark the fields with the #region format as used in the widely used code editor vs code.

An example of how a docstring could look like is the following:

```julia
@doc raw"""
    add_constraint_nodal_balance!(m::Model)

Balance equation for nodes.

#region formulation
\begin{aligned}
& v_{node\_injection}(n,s,t) \\
& + \sum_{\substack{(conn,n',d_{in},s,t) \in connection\_flow\_indices: \\ d_{out} == :to\_node}}
 v_{connection\_flow}(conn,n',d_{in},s,t)\\
& - \sum_{\substack{(conn,n',d_{out},s,t) \in connection\_flow\_indices: \\ d_{out} == :from\_node}}
 v_{connection\_flow}(conn,n',d_{out},s,t)\\
 & + v_{node\_slack\_pos}(n,s,t) \\
 & - v_{node\_slack\_neg}(n,s,t) \\
& \{>=,==,<=\} \\
& 0 \\
& \forall (n,s,t) \in node\_stochastic\_time\_indices: \\
& p_{balance\_type}(n) != balance\_type\_none \\
& \nexists ng \in groups(n) : balance\_type\_group \\
\end{aligned}
#endregion formulation

#region description
### [Nodal balance](@id constraint_nodal_balance)
In **SpineOpt**, [node](@ref) is the place where an energy balance is enforced. As universal aggregators,
they are the glue that brings all components of the energy system together. An energy balance is created for each [node](@ref) for all `node_stochastic_time_indices`, unless the [balance\_type](@ref) parameter of the node takes the value [balance\_type\_none](@ref balance_type_list) or if the node in question is a member of a node group, for which the [balance\_type](@ref) is [balance\_type\_group](@ref balance_type_list). The parameter [nodal\_balance\_sense](@ref) defaults to equality, but can be changed to allow overproduction ([nodal\_balance\_sense](@ref) [`>=`](@ref constraint_sense_list)) or underproduction ([nodal\_balance\_sense](@ref) [`<=`](@ref constraint_sense_list)).

The constraint consists of the [node injections](@ref constraint_node_injection), the net [connection\_flow](@ref)s and [node slack variables](@ref Variables).
#endregion description
"""
function add_constraint_nodal_balance!(m::Model)
    [...]
end
```

Summary of design decisions:
1. Latex in the docstring
    a. If we want to directly write latex equations we have to add the Documenter dependency to SpineOpt and use its docstrings (@doc raw""" """)
    b. If we want to keep normal docstrings, we'll have to correctly escape the special latex characters
2. Approach
    a. Use the docstring directly
        i. should be rather straightforward to implement
        ii. the content of the docstring will either deviate from the standard approach or will look very artificial in the documentation
    b. Find specific fields in the docstring
        i. the implementation requires a more serious script
        ii. flexible way to create the documentation with multiple fields: #region formulation ; #region description (#region used because the widely used vs code can recognise that field too)

tarskul:
An optional feature to pull documentation from a docstring. The intended use is a list of instructions that refers to fields in the docstrings.

The 'instruction file' is simply a regular markdown text but with the occasional 'instruction' encapsulated in `#region instruction` `#endregion instruction`. The first line in the instruction refers to the function from which you want to pull the documentation and the other lines refer to the specific fields from the docstring that are of interest to you.

The docstring itself starts now with @doc raw to allow for regular latex in the docstring (which adds a dependency on documenter). The fields are also encapsulated `#region fieldname` `#endregion fieldname`.

Fixes #649 

## Checklist before merging
- [x] Unit tests pass


Tasqu (approved):
Looks good to go as far as I'm concerned. @tarskul can you handle the merge?

tarskul:
@Tasqu Nope, I'm not authorized to merge

PR #778, opened 2023-10-19
Add reserve tutorial and figures

### LINKED ISSUE #751: Create a tutorial to define reserves constraints
**Is your feature request related to a problem? Please describe.**
Create a new tutorial to define ramping and reserves in SpineOpt

**Describe the solution you'd like**
Add a new file in the tutorial section of the documentation with the steps to define reserves constraints

**Describe alternatives you've considered**
N/A

**Additional context**
N/A

datejada:
(description of changes)

Fixes #751 

## Checklist before merging
- [NA] Unit tests pass


gnawin:
In this tutorial, we are going to add a ramping example as well.

tarskul:
Maybe a small remark on adding ramping to the tutorial. You can try to add it to the reserves tutorial but if it would become too complicated then perhaps we can consider to first make a ramping tutorial without reserves (starting from the simple system) and then later have a separate tutorial that combines the ramping and reserves tutorial?

datejada:
@tarskul, we were discussing something similar; the thing is that to define the reserves in SpineOpt you need to define the ramping parameters (as it is today). So, it might be better to put them together. Nevertheless, we can make it a different tutorial if it gets too complicated or lengthy. We will be sure to keep you posted.

datejada:
@tarskul, after looking into the details, we can actually define the reserves without defining the ramping parameters. I have updated the tutorial for ramps and made a new one for the ramps since they are not linked and can be independent. Sorry for the confusion before.

manuelma:
@datejada thank you for this, very nice work.

I still have one question though. Why do we need the node group? Is it possible to clarify the rationale behind it in the text? And actually, is the group really needed? Does the behaviour change if the group is not created? Do results change?

PR #784, opened 2023-10-30
Add tutorial for ramping constraints

### LINKED ISSUE #783: Create a tutorial to use the ramping constraints
**Is your feature request related to a problem? Please describe.**
Create a tutorial to define step-by-step the ramping constraints in SpineOpt

**Describe the solution you'd like**
New markdown file in the documentation

**Describe alternatives you've considered**
N/A

**Additional context**
N/A

datejada:
(description of changes)

Fixes #783 

## Checklist before merging
- [NA] Unit tests pass


datejada:
@tarskul, this tutorial is also ready for your review 😉 

@nnhjy, your feedback is also welcomed since you also have worked with this feature in SpineOpt

tarskul:
What do you mean with 'the model must finish in the second hour to properly account for the first hour'?

tarskul:
Maybe instead of posting my remarks one by one I should give you an overview:

- I like the overall work
- there are two parts in the beginning that were a bit confusing if I was *pretending* to be someone in a rush who needs to do this in between meeting
    - 'for the demand we need to change the model end and start so let's start with temporal' That is how I read it but I know that is not how it is there. Maybe just an indication that it could be reprhased to be a bit smoother? Maybe I'll try to come up with something myself.
    - What do you mean with 'the model must finish in the second hour to properly account for the first hour'? Maybe move this to a !!! note ?
- you say that SpineOpt uses the limit and the max ramp but you don't say why so that would be nice to add
- It is nice that you also include the initial condition

manuelma:
@datejada we need to coordinate I think, because I'm closing to finishing https://github.com/spine-tools/SpineOpt.jl/pull/789 and when that's merged this one needs to be adapted. How should we proceed?

datejada:
> @datejada we need to coordinate I think, because I'm closing to finishing #789 and when that's merged this one needs to be adapted. How should we proceed?

I think you should go first, then I change the tutorial (although it should be just clarifying that older versions need the extra parameters that won't be in the future versions, since there might be people using older versions? I don't know

manuelma:
> > @datejada we need to coordinate I think, because I'm closing to finishing #789 and when that's merged this one needs to be adapted. How should we proceed?
> 
> I think you should go first, then I change the tutorial (although it should be just clarifying that older versions need the extra parameters that won't be in the future versions, since there might be people using older versions? I don't know

Good point. So far our policy is to never include documentation for older versions and just upgrade DBs automatically to the latest version so we don't need to worry about any of that. My suggestion is to keep doing the same, just assume everybody has the latest version of SpineOpt and the data.

clizbe:
Waiting for the changes in the reserves and ramping constraints.

PR #786, opened 2023-10-31
WIP: Change constraint unit capacity

gnawin:
This is a draft PR about constraint unit capacity. 

Capture

It is part of #654 

## Checklist before merging
- [x] Unit tests pass


gnawin:
> Thank you @gnawin
> 
> So I have an idea to generate the constraint indices. I think what we need to do is split the stochastic path into continuous chunks where the check (min_up_time > delta) has the same outcome.
> 
> For example, say the stochastic path is equal to [s1, s2, s3, s4, s5] (where s1, s2, s3, s4, s5 are just scenarios).
> 
> Now say (min_up_time > delta) is true for scenarios s1, s2, and s5, but false for scenarios s3 and s4. Then we would split the original path into three:
> 
> * [s1, s2]
> * [s3, s4]
> * [s5]
> 
> For the subpaths [s1, s2] and [s5] we would generate the first constraint, and for the subpath [s3, s4] we would generate the other two constraints. In general, min_up_time doesn't change across scenarios so we would be using the original path unsplitted most of the time. If you agree, I can give it a go now directly in this branch?

Thanks!
So if I understood correctly, the check (min_up_time > delta) may differ per scenario, so for different scenarios, we may need to activate different constraints. That makes sense to me.
Implementation-wise, I guess it may mean we need to somehow split the path, allocate the scenarios to different constraints, and let the constraints somehow depend only on the scenarios instead of the whole path. This is not yet clear to me how you would do it, but would be great if you can give it a try!

manuelma:
> > Thank you @gnawin
> > So I have an idea to generate the constraint indices. I think what we need to do is split the stochastic path into continuous chunks where the check (min_up_time > delta) has the same outcome.
> > For example, say the stochastic path is equal to [s1, s2, s3, s4, s5] (where s1, s2, s3, s4, s5 are just scenarios).
> > Now say (min_up_time > delta) is true for scenarios s1, s2, and s5, but false for scenarios s3 and s4. Then we would split the original path into three:
> > 
> > * [s1, s2]
> > * [s3, s4]
> > * [s5]
> > 
> > For the subpaths [s1, s2] and [s5] we would generate the first constraint, and for the subpath [s3, s4] we would generate the other two constraints. In general, min_up_time doesn't change across scenarios so we would be using the original path unsplitted most of the time. If you agree, I can give it a go now directly in this branch?
> 
> Thanks! So if I understood correctly, the check (min_up_time > delta) may differ per scenario, so for different scenarios, we may need to activate different constraints. That makes sense to me. Implementation-wise, I guess it may mean we need to somehow split the path, allocate the scenarios to different constraints, and let the constraints somehow depend only on the scenarios instead of the whole path. This is not yet clear to me how you would do it, but would be great if you can give it a try!

Exactly, min_up_time may be different (in principle) for different scenarios. So if we use the full stochastic path, that has many scenarios, in our constraint index, we cannot know for certain in which case we are with respect to (min_up_time > delta).

The easiest solution would be to split the path into as many subpaths as scenarios, but that could be overkill. My solution is to split the path into chunks that'd lead to different constraints, and in the best-case (and more frequent) *scenario*, we would end up using the full path.

manuelma:
I just push my idea @gnawin 

I didn't test it but I was hoping you could somehow give it a look and try to validate it?

One thing that I'm pretty sure won't work at the moment is the `max` operation for the second and third coefficients on the rhs. We need to make it work with any type of parameters (say, stochastic time series) and for that we need to develop SpineInterface. I can do that next.

nnhjy:
I have a stupid question: 

In equation 1 with only $u_t$ and $y_t$ being 1, the flow $f_t$ represents a startup flow as being bounded by startup capacity (correct me if I'm wrong). However, it is supposed to denote a unit flow in some other cases. How does the new implementation distinguish this role disparity of $f_t$ in SpineOpt?

manuelma:
> I have a stupid question:
> 
> In equation 1 with only ut and yt being 1, the flow ft represents a startup flow as being bounded by startup capacity (correct me if I'm wrong). However, it is supposed to denote a unit flow in some other cases. How does the new implementation distinguish this role disparity of ft in SpineOpt?

The way I understand it, these three constraints will cover multiple cases. The one you mention where `u == y == 1` results in `flow <= startup_limit` - which makes sense because the unit is starting up now. But for example if `u == 1` and `y == 0` (the unit already started up in a previous time step), then the constraint becomes `flow <= capacity` which also makes sense because now the unit is just operating.

manuelma:
@gnawin I have pushed some more commits and now it looks like the new constraint is working alright. All tests that use the constraint pass, but of course the test that tests the constraint directly fails because the indices are different etc. Maybe you can try and fix that test and then we can merge?

nnhjy:
Thanks @manuelma. I seem to miss something. Does $u_t$ mean the number of online units at time `t`? If so and $y_t$ means start-up units at `t`, then at that very moment the unit flow should consist of one from the online and one from started-up, rather than merely the startup flow. If not, what does $u_t$ here stand for?

manuelma:
> Thanks @manuelma. I seem to miss something. Does ut mean the number of online units at time `t`? If so and yt means start-up units at `t`, then at that very moment the unit flow should consist of one from the online and one from started-up, rather than merely the startup flow. If not, what does ut here stand for?

If you replace `u_t = 1, z_t+1 = 0, u_t^N- = 0, y_t = 1` on the rhs you end up with something like `F - (F - SU)` which is equal to `SU`. So you are in the case where the unit is starting up and not shutting down in the next period, and in that case the flow should be lower than SU. So it is correct. You can do the same with all the other combinations of values for `u_t`, `y_t` etc and find the different cases.

nnhjy:
Ah I see. So `u_t == y_t ==1` means the unit online at time `t` is starting up. If `u_t == 2` with `y_t ==1`, then there is 1 unit online and the other starting up at time `t`, with rhs being `2F - (F - SU)*1`.

datejada:
> Ah I see. So `u_t == y_t ==1` means the unit online at time `t` is starting up. If `u_t == 2` with `y_t ==1`, then there is 1 unit online and the other starting up at time `t`, with rhs being `2F - (F - SU)*1`.

@nnhjy correct, if you factor the values, you will get that flow <= F+SU, which means that since one is online and the other is starting up, then the total flow can't be greater than F (full capacity flow of the unit online) + SU (the startup capacity of the one starting up).

This is a clustered unit commitment, where the unit commitment variables can be integer. I can recommend the following references for that topic if you want to dig in (one is a work from German and me 😉):

https://ieeexplore.ieee.org/document/8003295

https://ieeexplore.ieee.org/document/8676091

gnawin:
> then the total flow can't be

Thanks a lot Manuel. I will start to look at your changes tomorrow and try to fix that particular test.

manuelma:
Hi @gnawin 

I think there is a small mistake in the constraint still, I will fix it now.

gnawin:
Hi @manuel,
Thanks for the work! It looks great. I have one comment though, for the condition regarding the `duration(t)`, it should be about `duration(t_on)` instead of `duration(t_flow)`. That concerns line 148 to 150, I propose to replace `t_flow` with `t_on = t_overlaps_t(m; t=t_flow)`, i.e., :
```
  t_on = t_overlaps_t(m; t=t_flow)
  t_on_duration = end_(t_on) - start(t_on)
  for s in path
      mut = min_up_time(unit=u, analysis_time=t0, stochastic_scenario=s, t=t_on, _default=nothing)
``` 
Would the above work?

manuelma:
> Hi @manuel, Thanks for the work! It looks great. I have one comment though, for the condition regarding the `duration(t)`, it should be about `duration(t_on)` instead of `duration(t_flow)`. That concerns line 148 to 150, I propose to replace `t_flow` with `t_on = t_overlaps_t(m; t=t_flow)`, i.e., :
> 
> ```
>   t_on = t_overlaps_t(m; t=t_flow)
>   t_on_duration = end_(t_on) - start(t_on)
>   for s in path
>       mut = min_up_time(unit=u, analysis_time=t0, stochastic_scenario=s, t=t_on, _default=nothing)
> ```
> 
> Would the above work?

Thank you @gnawin !

Remember `t_overlaps_t` returns an `Array` of `TimeSlice`, not a single one. I am not sure how would we chose the `t_on` among these. Me feeling is we should be ok if we just bypass this detail?

gnawin:
> > Hi @manuel, Thanks for the work! It looks great. I have one comment though, for the condition regarding the `duration(t)`, it should be about `duration(t_on)` instead of `duration(t_flow)`. That concerns line 148 to 150, I propose to replace `t_flow` with `t_on = t_overlaps_t(m; t=t_flow)`, i.e., :
> > ```
> >   t_on = t_overlaps_t(m; t=t_flow)
> >   t_on_duration = end_(t_on) - start(t_on)
> >   for s in path
> >       mut = min_up_time(unit=u, analysis_time=t0, stochastic_scenario=s, t=t_on, _default=nothing)
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Would the above work?
> 
> Thank you @gnawin !
> 
> Remember `t_overlaps_t` returns an `Array` of `TimeSlice`, not a single one. I am not sure how would we chose the `t_on` among these. Me feeling is we should be ok if we just bypass this detail?

Yes, I feel the same. But anyhow we need to select a `TimeSlice`, so like the following?

  t_on = t_overlaps_t(m; t=t_flow)
  t_on_duration = end_(first(t_on)) - start(first(t_on))
  for s in path
      mut = min_up_time(unit=u, analysis_time=t0, stochastic_scenario=s, t=t_on, _default=nothing)

manuelma:
> 
> Yes, I feel the same. But anyhow we need to select a `TimeSlice`, so like the following?
> 
> ```
>   t_on = first(t_overlaps_t(m; t=t_flow))
>   t_on_duration = end_(t_on) - start(t_on)
>   for s in path
>       mut = min_up_time(unit=u, analysis_time=t0, stochastic_scenario=s, t=t_on, _default=nothing)
> ```

That could work, very nice.

gnawin:
Just had a call with @mihlema regarding this constraint. One issue discussed is what happens to this constraint if no ramping/reserve parameters are defined.

1. I see some parameters are defined in this way, e.g., `max_startup_ramp[(...,_default=1)] `. Does this make sure that 1) this code still works in such a case, and 2) that only the basic version will be used?
2. maybe it can be a consideration to create this constraint or the basic one (as in current SpineOpt) based on methods, i.e., whether ramping/reserve parameters are defined.

The first one is a check for the current implementation, while the second one is a more high-level discussion.

manuelma:
> 1. I see some parameters are defined in this way, e.g., `max_startup_ramp[(...,_default=1)] `. Does this make sure that 1) this code still works in such a case, and 2) that only the basic version will be used?

Yes, if max_startup_ramp is not specified then it will use 1 as its value. That works. We have other tests that do not define max_startup_ramp and they pass, which means the constraint is being correctly generated for that case. I guess you could test the constraint with a given value of max_startup_ramp and see what happens?

> 2. maybe it can be a consideration to create this constraint or the basic one (as in current SpineOpt) based on methods, i.e., whether ramping/reserve parameters are defined.

That has advantages and disadvantages in my opinion. The disadvantage is I need to specify two parameters to trigger the behaviour (the start up limit itself, and the method to say 'use the limit that I am giving'). That can become quite annoying quite quickly. The advantage is I can try two versions of my model by switching the method, and I don't need to set my parameter value to nothing. But I'm pretty sure that's outside of the scope of this PR; let's forget about it for now, and when the time comes we can introduce methods everywhere if we want.

gnawin:
Hi @manuelma,

I had a discussion with @datejada @g-moralesespana. Unfortunately, there are some complications of this constraint (in SpineOpt terms) that were not clear before. 

See the updated picture at the beginning of this post.
 
1. Regarding the `t` indice of this constraint, we should always write for the highest resolution between `t_on` and `t_flow`.
 
2. Regarding the delta `t`, it is about `t_on`, but should be about its periods, not its duration value.
 
For example, for 6h, we have three different t_on resolutions [1h, 2h, 3h], called [Period1, Period2, Period3]. The situation, where units on variable `u_t = [0, 1, 0]`, is critical. And we want to compare TU with the difference of the number of periods, i.e., for the first two periods `u_2 - u_1`. So the duration of `u_t`, in this case `duration(t_1) = 1, duration(t_2) = 2, duration(t_t) = 3`, does not matter.

We might need a call to discuss further.

manuelma:
> Hi @manuelma,
> 
> I had a discussion with @datejada @g-moralesespana. Unfortunately, there are some complications of this constraint (in SpineOpt terms) that were not clear before.
> 
> See the updated picture at the beginning of this post.
> 
> 1. Regarding the `t` indice of this constraint, we should always write for the highest resolution between `t_on` and `t_flow`.
> 2. Regarding the delta `t`, it is about `t_on`, but should be about its periods, not its duration value.
> 
> For example, for 6h, we have three different `t_on` resolutions `[1h, 2h, 3h]`, called `[Period1, Period2, Period3]`. The situation, where units on variable `u_t = [0, 1, 0]`, is critical. And we want to compare TU with the difference of the periods, e.g., for the first two periods `u_2 - u_1 = 1`. So the duration of `u_t`, in this case `duration(t_u1) = 1, duration(t_u2) = 2, duration(t_u3) = 3`, does not matter.
> 
> We might need a call to discuss/clarify further.

@gnawin @g-moralesespana @datejada just trying to understand this. Can you please point out to what have changed in the original image?

Point 1 seems ok and I think I can handle it - but might require some work, trial and error.
Point 2 is a bit more confusing. I understand it is about distinguishing between the two cases we can have (TU == 1, TU != 1) but I don't understand how exactly it should work. What should we be comparing the minimum up time with? If you could just clarify this I should be able to write the code I think.

gnawin:
> > Hi @manuelma,
> > I had a discussion with @datejada @g-moralesespana. Unfortunately, there are some complications of this constraint (in SpineOpt terms) that were not clear before.
> > See the updated picture at the beginning of this post.
> > 
> > 1. Regarding the `t` indice of this constraint, we should always write for the highest resolution between `t_on` and `t_flow`.
> > 2. Regarding the delta `t`, it is about `t_on`, but should be about its periods, not its duration value.
> > 
> > For example, for 6h, we have three different `t_on` resolutions `[1h, 2h, 3h]`, called `[Period1, Period2, Period3]`. The situation, where units on variable `u_t = [0, 1, 0]`, is critical. And we want to compare TU with the difference of the periods, e.g., for the first two periods `u_2 - u_1 = 1`. So the duration of `u_t`, in this case `duration(t_u1) = 1, duration(t_u2) = 2, duration(t_u3) = 3`, does not matter.
> > We might need a call to discuss/clarify further.
> 
> @gnawin @g-moralesespana @datejada just trying to understand this. Can you please point out to what have changed in the original image?
> 
> Point 1 seems ok and I think I can handle it - but might require some work, trial and error. Point 2 is a bit more confusing. I understand it is about distinguishing between the two cases we can have (TU == 1, TU != 1) but I don't understand how exactly it should work. What should we be comparing the minimum up time with? If you could just clarify this I should be able to write the code I think.

I updated the picture again to minimize confusion.

Point 1 still holds, now I explain more about point 2. 

Yes, when everything is in hourly resolution, the distinction should be (TU == 1, TU != 1).

The rationale for this distinction is to avoid cases when units are online for only 1 time step for units_on (time step may not be the best name for this; I think in SpineOpt it is actually timeslice? Anyway, I try to explain what I mean). We only care about the resolution for units_on for the comparison of the TU.

The following may happen if minimum up time of a unit is 1, i.e., `TU == 1`
```
u_t = [0, 1, 0]
y_t = [0, 1, 0]
z_t = [0, 0, 1]
```
When `t = 2`, we have `u_2 = 1`, `y_2 =1`, `z_3=1`. The first constraint becomes `f <= F - (F - SD) - (F- SU)`, i.e., `f <= SD - F + SU`, which can be infeasible.


Moreover, since we just want to avoid this situation 
```
u_t = [0, 1, 0]
y_t = [0, 1, 0]
z_t = [0, 0, 1]
```
the duration, e.g., the hours, within a time step doesn't matter.

In summary, we are comparing minimum up time with 1 time step. Implementation-wise, still (TU > 1, TU <= 1). For TU, even it is is given in durations, we should somehow convert it. For the 1 time step, it is not measured in durations.

If the above is still not clear, let's arrange a call.

manuelma:
Thank you very much @gnawin !
It is clear now, will implement.

gnawin:
Hi @manuelma (also @datejada @g-moralesespana so you can follow the discussion and jump in if desired)

I am reading the code and want to get some clarification from you. 

```
t_flow_duration = end_(t_on) - start(t_on)
for s in path
    mut = min_up_time(unit=u, analysis_time=t0, stochastic_scenario=s, t=t_on, _default=nothing)
    mut_gt_dur = mut === nothing || mut > t_flow_duration
```

For every `t_on`, you get its duration, let's say `t_flow_duration = 2`.
 
Then (for every subpath) you get `mut` based on `t_on`, i.e., `mut` is directly in periods/timeslices of `t_on`.   

Therefore, this condition `TU > 1` in the constraint is equivalent to `mut` > `t_flow_duration`
but the value of `t_flow_duration` does not matter, this corresponds to what I said before

> the duration, e.g., the hours, within a time step doesn't matter. 

Did I understand the code/intention correctly with what I describe above?
 
But I have one more question, in SpineOpt, you can actually call `min_up_time` with different `timeslices`, in this case with `t_on`. In SpineToolbox, you have to define `min_up_time` with units, e.g., 8h. You can also define unit_on resolution as 30 min. But in our constraint, we cannot directly compare 8 with 30. Is there a conversion somewhere to make sure that they are measured in the same units in SpineOpt?

manuelma:
Hi @gnawin - are you looking at this branch? https://github.com/spine-tools/SpineOpt.jl/pull/799

That's the branch that has the newest version.

gnawin:
Hi @manuelma, thank you! I am now reading the latest master (#799 has the latest updates on this constraint), and still have some questions though 😅 . Since it is still about this constraint, I will continue the discussion here.

Great work! It seems to me there are lots of effects on implementation of the temporal structure and stochastics. 
Now I am trying to understand the code better to avoid anything misunderstandings.

Could you answer my questions below? Sorry for these detailed questions, I just could not figure it out at the moment and I think it is beneficial if we discuss there instead of me poking around. @datejada @g-moralesespana feel free to jump in.
 
- Q1. now `t` is from the highest `t` of `unit_on` and `unit_flow`, from below indices

```
for t in t_highest_resolution(
    Iterators.flatten(
        ((t for (u, t) in unit_time_indices(m; unit=u)), (t for (ng, t) in node_time_indices(m; node=ng)))
    )
)
```

Great! But I got confused by the use of `overlap_duration` and `t_overlaps_t`. I know their definitions, but I couldn't get it if that will give our desired outputs or not.

Referring to the below two parts
```
expr_sum(
    unit_flow[u, n, d, s, t_over] * overlap_duration(t_over, t)
    for (u, n, d, s, t_over) in unit_flow_indices(
        m; unit=u, node=ng, direction=d, stochastic_scenario=s, t=t_overlaps_t(m; t=t)
    ) 
```
```
+ expr_sum(
    _unit_flow_capacity(u, ng, d, s, t0, t) * units_on[u, s, t_over] * overlap_duration(t_over, t)
    for (u, s, t_over) in units_on_indices(m; unit=u, stochastic_scenario=s, t=t_overlaps_t(m; t=t));
    init=0
)
```

For example, if `unit_flow` is two-hourly and `unit_on` is hourly (btw, this is #797 talks about). The constraint should be written in hourly, i.e., something like `unit_flow * 1/2 <= unit_on * 1`. Is that what the constraint will generate?

Regarding the part on the comparison of `min_up_time` and 1.

- Q2: about units of `min_up_time`, by using `t` as an argument, i.e., `min_up_time(...t=t...)`, does this mean that `min_up_time` will be measured in the same unit as `t`? E.g., if `t` is in minutes, then `min_up_time` will also be in minutes. But in `SpineToolbox`, maybe I am wrong, but user can specify the units of `min_up_time`, would there be a conflict?

- Q3: `min_up_time` is only for `unit_on`, right? but in this `function _unit_capacity_constraint_subpaths(path, u, t0, t)`, in principle, `t` can also be from unit_flow, right?

- Q4: now suppose here `t` is for `unit_on`, so `min_up_time` is also for `unit_on`, great. We are only comparing the `min_up_time` and its `duration` in terms of `timeslice`, right? E.g., assume `unit_on` is in two-hourly. Then we only check if `min_up_time` is `>` 2h or `<=` 2h, which should be equivalent to the checks in the original formulation ( `TU` is `>` 1 or `<=` 1).   

Thank you very much!

manuelma:
Hi @gnawin - very good questions - here are my answers.

Q1: Yes, something like that. Note that `unit_flow` is always flow per `model_duration`. Maybe take a look at the unit-tests I wrote. If you could double check or eventually write more tests that's always great.
Q2: They won't be in the same units necessarily but comparisons will work all the same because we use `Period` values.
Q3: Yes, but it doesn't matter. A parameter value doesn't need to match the resolution of any of the variables. Values can be queried for any t.
Q4: Yes, exactly.

PR #798, opened 2023-11-14
780 temporal resolution tutorial

### LINKED ISSUE #780: Create a tutorial for flexible temporal resolution
Create a tutorial that starts with the Simple System and shows how to add:

- [x] Uniform temporal resolution
- [x] Varying temporal resolution by time
- [x] Varying temporal resolution by part of the system

This has a good demo for reference:
https://www.youtube.com/watch?v=FiiqZNcx7Ds&t=2484s

clizbe:
Created tutorial with photos of how to create:
- Uniform temporal resolution
- Varying temporal resolution by time
- Varying temporal resolution by part of the system

Fixes #780 

## Checklist before merging
- [NA] Documentation is up-to-date
- [NA] Unit tests pass

### DIFF
--- a/docs/src/tutorial/temporal_resolution.md
+++ b/docs/src/tutorial/temporal_resolution.md
@@ -2,29 +2,15 @@
 
 Welcome to Spine Toolbox's Temporal Resolution tutorial.
 
-This tutorial provides a step-by-step guide to include uniform and/or flexible temporal resolution in your model.
+This tutorial provides a step-by-step guide to include uniform and/or variable temporal resolution in your model.
 
 For more information on how time works in SpineOpt, see the [Temporal Framework](https://spine-tools.github.io/SpineOpt.jl/latest/advanced_concepts/temporal_framework/) documentation.
 
 ## Introduction
 
-### Installation and upgrades
-
-If you haven't yet installed the tools yet, please follow the installation guides: 
-- For Spine Toolbox: [Spine Toolbox installation guide](https://github.com/spine-tools/SpineOpt.jl#installation)
-- For SpineOpt: [SpineOpt installation guide](https://github.com/spine-tools/Spine-Toolbox#installation)
-
-If you are not sure whether you have the latest version, please upgrade to ensure compatibility with this guide.
-
-- For Spine Toolbox:
-    - If installed with pipx, then use `python -m pipx upgrade spinetoolbox`
-    - If installed from sources using git, then `git pull`, `python -m pip install -U -r requirements.txt`
-
-- For SpineOpt: [SpineOpt upgrade guide](https://github.com/spine-tools/SpineOpt.jl#upgrading)
-
 ### Model assumptions 
 
-- This tutorial builds upon the Simple System tutorial. Please follow [that tutorial](https://spine-tools.github.io/SpineOpt.jl/latest/tutorial/simple_system/) first, or download the [finished system](**NEEDS LINK**).
+- This tutorial builds upon the Simple System tutorial. Please follow [that tutorial](https://spine-tools.github.io/SpineOpt.jl/latest/tutorial/simple_system/) first, or download the [finished model](**NEEDS LINK**).
 - The scenario runs for 5 hours and the default temporal resolution is 1 hour.
 - The fuel_node has a temporal resolution of [1, 1, 1, 2] hours.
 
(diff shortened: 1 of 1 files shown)


docs/src/tutorial/temporal_resolution.md
```diff
@@ -0,0 +1,175 @@
...
+- Add model_end\
+	`parameter_name: model_end`\
+	`alternative_name: Base`\
+	`value:` Edit > `Parameter Type: Date Time` > 5 hours after start
+
+Your *Object parameter value* window show look like this:
+
+![image](figs_temporal_resolution/temporal_resolution_model_start_end.PNG)
+
+!!! Regularly save your changes by selecting the *Hamburger menu* (top right) > Commit > Write a message, such as "Save" > OK
```
tarskul, line 83: Do you intend to write a note? In that case the keyword 'note' may be missing...

docs/src/tutorial/temporal_resolution.md
```diff
@@ -0,0 +1,175 @@
...
+You should see different colored lines, for the *electricity* and *fuel* flows that are *to* and *from* the powerplants (red and green are on top of each other). The blue line shows the electricity demand of [10, 20, 40, 50, 20] - and how the fuel demand from Powerplant A is this same array divided by 0.7 (the conversion ratio).
+
+![image](figs_temporal_resolution/temporal_resolution_uniform_results.PNG)
+
+!!! If `Time series` entries are not visible, select the Hamburger menu > View > Pivot Table: Value
+
+### Flexible temporal resolution
+
+#### Creating a time-varying resolution
+
```
tarskul, line 126: It is a bit dry if you just say do this do that, perhaps give it a bit of flavour by telling why we would want to do this?

tarskul, line 126: The main downside of this approach is that we do not have an intermediate result that we can run. Maybe you can tell in a note that an interested user may temporarily switch the default from flat to non flat to see the difference in results?

docs/src/tutorial/temporal_resolution.md
```diff
@@ -0,0 +1,175 @@
...
+- In the *Object parameter value* window, find the value for the electricity node's demand (should be 150 from the Simple System) - right click > Edit\
+	`Parameter type: Time Series fixed resolution`\
+	Copy and paste (or fill in the table):
+	
+		10
+		20
+		40
+		50
+		20
+
```
tarskul, line 55: I don't know whether it matters but with this (low) demand you effectively disable unit b.

tarskul (commented):
Nice work! The experience was smooth as butter.

I just left a few remarks for further improvements.

docs/src/tutorial/temporal_resolution.md
```diff
@@ -0,0 +1,175 @@
...
+- The scenario runs for 5 hours and the default temporal resolution is 1 hour.
+- The fuel_node has a temporal resolution of [1, 1, 1, 2] hours.
+
+## Tutorial
+
+### Adjustments to Simple System
+
+Before demonstrating flexible temporal resolution, we need to adjust the Simple System tutorial so the results are more interesting to examine.
+
+!!! note If you are ever missing a window in the Spine DB Editor, go to the *Hamburger menu* (top right) > under View: Docks > Choose the missing window - then drag and resize the window to your preference
```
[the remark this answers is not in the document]
clizbe: Ha yes! I'm not exactly sure what the default view is, so referencing something that might not be there is an easy way to lose the noobs who are trying to follow along.

docs/src/tutorial/temporal_resolution.md
```diff
@@ -0,0 +1,175 @@
...
+
+#### Adding variable electricity demand
+
+In SpineToolbox, double click on the *InputData*  block.
```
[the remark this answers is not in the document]
clizbe: Yes I talked to Diego about this. It depends whether we give them the entire project or just the JSON. If it's the JSON I'll need to add a bit about setting up the workflow and importing the JSON.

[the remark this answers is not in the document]
tarskul: We'll provide the JSON file as that is more github friendly. The reason why I haven't put them on there yet is because I'm also coupling unit tests to the JSON files but I can't get the tests to work .... yet.

docs/src/tutorial/temporal_resolution.md
clizbe: Good catch.

docs/src/tutorial/temporal_resolution.md
clizbe: Yeah I combined creating a time-changing resolution and assigning a resolution to a specific part of the model. That's a good idea to make it the default just to show that's also an option.

docs/src/tutorial/temporal_resolution.md
clizbe: Ha yeah. The graph is quite boring. Do you think it matters enough to change all the graphs?

docs/src/tutorial/temporal_resolution.md
clizbe: That's what I was thinking too.

docs/src/tutorial/temporal_resolution.md
tarskul: Not sure, there are advantages and disadvantages in doing so. Now it is simpler and therefore easier to follow. And it may be good that there are a few easy tutorials for confidence boosting. On the other hand, it could be interesting to compare two similar units with different resolutions (though that would require us to introduce another node which may introduce more errors).

Perhaps we give a hint that this is something that they can try to do themselves but which is not covered in the tutorial (except perhaps one graph with the solution such that they know what to aim for)?

docs/src/tutorial/temporal_resolution.md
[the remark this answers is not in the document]
clizbe: I updated the instructions to work with a JSON file.

docs/src/tutorial/temporal_resolution.md
clizbe: I'll leave it simple as it is. It's more about HOW to put in resolution, not what cases and why. They should know that! ;)

docs/src/tutorial/temporal_resolution.md
clizbe: Added this

clizbe: Although I'm not really sure what to say. I added a tiiiiny bit of flavor. Maybe it'll help but let me know if it's still boring.

Issue #602, opened 2023-04-12
Explain PTDF-based investments are correct for only 1 line


jkiviluo:
Explain that PTDF based investments are correct only for one line. If there are more and those interact, then reactances (?) will start to go wrong and the results become increasingly more incorrect. Should be explained in the documentation to make users aware. And mention that the angle based investments work anyway but is slower (and won't have N-1).

clizbe:
@tarskul If you understand this, would you mind adding it to the docs? You're good at explaining things. :)

tarskul:
@clizbe I'm not too familiar with PTDF in SpineOpt (and it has been a while since I last worked with PTDF in general) but I'm willing to learn and as I learn I can indeed document what I learned. However, I don't think that I will put this high on my priority list, so it may take a while before I get to it. In the meanwhile we can add a warning note.

PR #836, opened 2024-01-05
Add a warning that investments with the PTDF method only work for one line

tarskul:
Add a warning on investments with the PTDF method as discussed in #602


clizbe (approved):
LGTM!

docs/src/advanced_concepts/powerflow.md
DillonJ, line 1: @tarskul @clizbe it's not quite complete to say that the calculations are "only correct for one line". Strictly, the calculations are correct for one line at a time, which is different. For example, one may use the method to evaluate multiple optional investments that are mutually exclusive (chose one of N possible investments).

Perhaps it would be better to say something along the lines of "the calculations do not consider the mutual effect of multiple simultaneous investments, so this method is useful for choosing between mutiple simultaneous investments that are assumed non-interacting and/or multiple mutually exclusive investments.

Also, it should be relatively easy to extend the method to overcome this limitation - we should create an issue

tarskul: I've changed the text with your recommendation. Is it better this way?

Issue #735, opened 2023-08-18
Check if tutorials are working


clizbe:
Go through each of the tutorials and see if they still work and are followable.
If not, create issues to update them so they work.

- [ ] Tutorial 1
- [ ] Tutorial 2
- [ ] Tutorial 3

clizbe:
Related: #436, #459

tarskul:
I've updated the simple system tutorial with the experience from my own struggle, mostly explaining some things in more detail (e.g. the three-way relationship for defining an efficiency) and adding some warnings for pitfalls (e.g. messing up the order in the three-way relationship). Besides the pitfalls, the tutorial seems fine.

tarskul:
After some deliberation, only the first tutorial is a real tutorial. The other two should be adapted to examples or how to sections.

Then we should also add some new tutorials that cover more features of SpineOpt.

I'll edit the issue description and we should later have a discussion on these new tutorials.

datejada:
I'm working on the following tutorial: 

- [ ] Create a tutorial to define reserves

I created a new issue and branch for it

datejada:
The unit commitment tutorial is in #765

PR #891, opened 2024-02-20
accept dictionary as input for SpineOpt and use this feature to test the tutorials

tarskul:
Spine Toolbox can export and import to json, so why not let SpineOpt accept Julia dictionaries (that are potentially parsed from json files)?

This feature is also useful for testing as the test setups can be written directly in a human readable json file instead of using SpineInterface to create a spine db. Of course, under the hood SpineInterface is used anyway to change the dictionary to a spine db. But, for the user/developer it is seemingly easier to work with dictionaries.

In this branch, the feature is used to test the tutorials (#735 ).

However, there are two questions that remain:
+ Currently, it is done through the julia feature where you can re-use the same name but with different input types, is that the best way to go about this?
+ Currently, the tests use 'upgrade==true' which may cause issues for new users. So the question is how often do we want to update these tutorial files? Every time that they fail without upgrading? Or only when the upgrading feature does not work?

## Checklist before merging
- [ ] Documentation is up-to-date -> should this be in the developers documentation or should it also be communicated to the users?
- [x] Unit tests pass


src/run_spineopt.jl
```diff
@@ -171,6 +172,56 @@ function run_spineopt(
...
+    resume_file_path=nothing,
+    run_kernel=run_spineopt_kernel!,
+)
+    db_url = "sqlite://"
+    data_template = Dict(Symbol(key) => value for (key, value) in template())#SpineOpt.template()
+    data_in = Dict(Symbol(key) => value for (key, value) in url_in)
+    data_run = merge(data_template, data_in)
+    SpineInterface.close_connection(db_url)
+    SpineInterface.open_connection(db_url)
+    SpineInterface.import_data(db_url, "From Julia dictionary to Spine database."; data_run...)
```
manuelma, line 201: This requires PyCall to be properly configured, are we sure we want that?

`using_spinedb` already accepts a dictionary as an input. If you look at `prepare_spineopt`, you will see that we actually use that method. We call `data = export_data(url_in; upgrade=upgrade, filters=filters)` followed by `using_spinedb(data, @__MODULE__; extend=true)`. But if `url_in` is already a dict, then we could just replace the former with `data = url_in` and proceed.

manuelma: We don't need to define an entire new `run_spineopt` method. Maybe we should try that idea?

tarskul: I'm not really sure whether I fully understand all of this, but I'm certainly willing to try.

But can we really avoid the new run_spineopt? Do you want a check whether the input is a string or a dictionary and then run the one or the other? And would run_spineopt then take something like run_spineopt(url_in::Union{String, Dict})?

Tasqu: Sounds good to me

Tasqu (approved):
I'm happy with this if @manuelma is.

tarskul:
Ok, great! Let's do it in two steps then. I'll merge it now for the functionality and in a second step we can do code simplifications and consider a higher update frequency for the tutorial files.

I've tried the code simplification (and I think I got around the url_out by passing it to prepare spineopt and set it to nothing if the input is a dictionary, assuming that the functionality is used in a script and the user will handle the output later themselves) but I got to a point where I don't understand the error ('pybytes' error). After the merge I can push those changes to the 'acceptdictionary' branch so you can start from there @manuelma ? Or I can let it be and you can indeed refactor it to your own liking later.

manuelma:
Good job @tarskul - no worries, I can fix later.

PR #895, opened 2024-02-21
685 bidirectional connection becomes unbounded

### LINKED ISSUE #685: Bidirectional connection becomes unbounded
It seems to me that a bidirectional connection becomes unbounded with some common settings.

**To Reproduce**
For connection between node_a and node_b: Set `connection__to_node` and `connection__from_node` both for node_a and node_b. Set `fix_ratio_out_in_connection_flow` to both directions, at most 1 and one direction strictly smaller than 1. Set `connection_capacity` to some values. Set `balance_type_none` for node_b.

The connection is then free to consume flow from node_b, unlimited by its capacity.

**Expected behavior**
One would expect it to be limited by the capacity.

- Which SpineOpt version?  Master from Jun 15th.

nnhjy:
(description of changes)

Fixes #685 as per [this approach](https://github.com/spine-tools/SpineOpt.jl/issues/685#issuecomment-1929439455)

## Checklist before merging
- [x] Unit tests pass

### DIFF
--- a/src/constraints/constraint_connection_flow_capacity.jl
+++ b/src/constraints/constraint_connection_flow_capacity.jl
@@ -76,7 +76,16 @@ function add_constraint_connection_flow_capacity!(m::Model)
                 );
                 init=0,
             )
-            - connection_capacity[(connection=conn, node=ng, direction=d, stochastic_scenario=s, analysis_time=t0, t=t)]
+            + expr_sum(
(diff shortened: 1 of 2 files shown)


DillonJ:
I think we should instead implement a constraint/bount on the flows individually to support different capacities in each direction

PR #897, opened 2024-02-21
Remove PyCall dependency to run_spineopt with a Dict

manuelma:
Also upgrade examples json to latest structure.

## Checklist before merging
- [x] Unit tests pass


tarskul:
Cool stuff happening here. :) There are some parts that show me that I was thinking in the right direction, but there is also some stuff that I would've never guessed. Good to learn!

PR #918, opened 2024-03-05
WIP: Try to fix large benders gaps

manuelma:
We are seeing large gaps that stay near 200% for many iterations.

My theory is this happens because we mix together marginal values and subproblem objective values for all the rolling subproblems to create one big Benders cut. This leads the master problem to think that an investment can reduce the objective value in any window, even though the non-zero marginal value is only for some of the windows. So the master problem invests, reduces its objective function, but then the individual subproblems cannot materialize this reduction leading to a large gap.

What I am trying in this branch is to create one cut per rolling subproblem (instead of one big cut mixing all the subproblem together). Then, investments that only benefit some subproblems will indeed be counted as benefitting only those subproblems in the master problem objective function. So far it's proven beneficial to reduce the gaps, but I'm still a bit skeptical. Wouldn't this be a known issue in Benders decomposition? How come we didn't do it this way from the beginning?

@DillonJ any thoughts?

## Checklist before merging
- [x] Unit tests pass


manuelma:
After some more testing and fixing some bugs, it looks like the impact of having one cut per window is not that good. It might be that the large gaps we are seeing are just due to the model very slowly converging to the solution?

Anyways, I need to look more into it...

PR #922, opened 2024-03-05
685 bidirectional connection becomes unbounded

### LINKED ISSUE #685: Bidirectional connection becomes unbounded

nnhjy:
Implement tight linear formulation for the `connection_capacity` of `connection_flow`

The corresponding formulation for `connection_intact_flow` remains to change.

Fixes #685

## Checklist before merging
- [x] Unit tests pass


src/constraints/constraint_connection_flow_capacity.jl
```diff
@@ -67,37 +115,118 @@ function add_constraint_connection_flow_capacity!(m::Model)
...
-            <=
-            + connection_capacity[(connection=conn, node=ng, direction=d, stochastic_scenario=s, analysis_time=t0, t=t)]
-            * connection_availability_factor[(connection=conn, stochastic_scenario=s, analysis_time=t0, t=t)]
-            * connection_conv_cap_to_flow[
+        (
+            !isnothing(d_reverse)
+            #TODO: would using realize() significantly threaten the performance?
+            && realize(connection_capacity[
+                (connection=conn, node=ng, direction=d, stochastic_scenario=s, analysis_time=t0, t=t)
+            ]) * realize(connection_conv_cap_to_flow[
```
manuelma, line 123: We can't use realize, we need to keep coefficients as Call otherwise they don't update themselves. This would lead to unexpected results. I suggest we revert this PR and look at it in more detail.

manuelma: Actually I think it's pretty close - maybe no need to revert. But we should try and polish it as soon as possible.

nnhjy: Thanks for the comment! I also felt it not a good way to explicitly use `realize()` there. 

The problem of not using `realize()` is simply the lack of method `base.isless()` for a call and a number. Actually, we may need both `isless(x::Call, y::number/Int)` and/or `isless(x::Call, y::Constant)` if we prefer `Constant(0)` rather than 0.

nnhjy: Wouldn't the `realize()` evaluate its call to find the parameter value for the given indices?

At that place, the function need check whether the exact parameter value is strictly positive before deciding which constraints to create. So it seems that we cannot simply keep the call of the parameter as such, instead, it needs to be executed to find the corresponding parameter value w.r.t. the `(s, t0, t)` for the constraint.

manuelma: In other words, we cannot use the realized value of a Call to decide which constraint to build, because if the value changes in the next window, we cannot build the other constraint instead. All we can do is update coefficients and right-hand-sides of existing constraints. At the moment, we cannot deactivate one constraint and activate another - although that would be a very nice addition. I wonder if it's even possible.

nnhjy: Just tested it in a real run. It is actually working as expected.

manuelma: Do you mean it switches to the other constraint when the realized value of the Call changes? That would be very surprising...

manuelma: Have you tested a case where the capacity is a time-series that sometimes is zero and sometimes is not zero, and the model is rolling so in one window, the capacity would be zero and in the next window, the capacity would be greater than zero?

manuelma: One option would be to build a single constraint that would become one or the other depending on the realized value of the Calls, by carefully choosing the coefficients. In other words, try to put the 'tests' (`capacity[(...)] >= 0 ? coefficient if true : coefficient if false`) into the constraint coefficients directly. Would that work?

nnhjy: Right. That was also one candidate approach. I'll give it a try today.

manuelma: Would you consider reverting the commits to master in the meantime? It can help people avoid bugs...

nnhjy: Update on alternative approach @manuelma: 

0. To use `capacity[(...)] >= 0` for test during the creation of the constraint `isless(x::Call, y::Number)`. But then, what do we compare between a `Call` and a number? The call needs to be evaluated so that its value is compared with another number. This would need a method similar to the `realize()` for a `call` to fetch the corresponding parameter value during the creation of the constraint and before it gets evaluated.

1. As you mentioned above, this also relates to how a constraint is created and updated. To my understanding (correct me if I'm wrong), the constraints are made, and whether a constraint exists is fixed w.r.t. the first optimisation window before getting evaluated to obtain the coefficients' value during the solving (and rolling if it exists). The idea proposed here doesn't work with the current framework as it dynamically creates the constraint w.r.t. the running optimisation window. The constraint function needs to be called for each optimization window to generate the dedicated constraint. Otherwise, suppose a constraint is not generated at a point in the 1st window, but the parameter value in the same relative place of the next window asks for the use of the constraint. That would cause a problem.

2. The idea aims to generate minimal constraints as needed. And I remember there was a time @manuelma you implemented the current fancy coefficient updating mechanics, which greatly boosts the running speed on rolling : ). Is it possible and **realistic** to implement the dynamic constraint creation approach along with the current coefficient updating mechanism? There seems to be some benefit of applying a hybrid approach but also a lot to do : (.

manuelma: We could implement 'constraint switching' using the same mechanism as we do for coefficient and rhs updating, that's correct @nnhjy - but it's not so straightforward.

Issue #920, opened 2024-03-05
Add an expression for capacity_margin and a minimum capacity margin constraint to allow scheduling of outages


DillonJ:
With #903 we can now schedule maintenance outages. At present, feasibility and cost will be the only factors influencing the optimality of the scheduled outages. If we implement capacity margin as an expression, this allows adequacy to influence the decision. For example, we can define a minimum capacity margin that is penalised - this will force the model to schedule outages while respecting a minimum level of reliability and opens the door for more comprehensive treatments of adequacy.

clizbe:
@DillonJ Is this part of the updates you said you're working on in relation to another project?
In principal I think it's fine to add an expression (since it's not a variable), but in general I think we should focus on our major changes and leave adding smaller details like this (that also add constraints) until after we get the core and speed taken care of.

clizbe:
Closed by #924

PR #924, opened 2024-03-06
#920 capacity margin penalty

DillonJ:
Implements a minmum capacity margin constraint. The minimum capacity margin is calculated as an expression and introduces the concept of expressions which now appear alongside variables and constraints in the folder structure and tests.

Defining the `min_capacity_margin` parameter for a node triggers the creation of the `capacity_margin` expression and the `min_capacity_margin` constraint. The `capacity_margin` is essentially the difference between available unit capacity and demand at a node.

This is useful for scheduling units outages, preventing them from being scheduled at times of low capacity margin, and for triggering investments due to capacity shortage.

The constraint can be softened by specifying `min_capacity_margin_penalty` which triggers the creation of the `min_capacity_margin_slack` variable which is penalised with this coefficient in the objective function. This is also useful for disincentivising scheduled outages during times of low margin.

```math 
\begin{aligned}
expr^{capacity\_margin}_{n,s,t} = \\
& + \sum_{u\in{U_{n\_to}}}(p^{unit\_capacity}_{u,s,t} \cdot p^{unit\_availability\_factor}_{u,s,t} \cdot v^{units\_available}_{u,s,t}) \\
& + \sum_{u\in{U_{storage_n}}}(v^{unit\_flow}_{u,n,to,s,t}) \\
& - \sum_{u\in{U_{storage_n}}}(v^{unit\_flow}_{u,n,from,s,t}) \\
& - p^{demand}_{n,s,t} \\
& - p^{fractional\_demand}_{n,s,t} \cdot p^{group\_demand}_{n_{group},s,t} \\
& \forall n \in node: p^{min\_capacity\_margin} \\
\end{aligned}
```
where ```math U_{storage_n} ``` is the set of all storage units connected to node n
and ```math U_{n\_to} ``` is the set of all non-storage units connected to node n

Completes #920

## Checklist before merging
- [x] Unit tests pass


DillonJ:
@manuelma maybe we should be tighter with our words regarding when an MR can be merged or not :-)

manuelma:
> @manuelma maybe we should be tighter with our words regarding when an MR can be merged or not :-)

Sure @DillonJ - this time I'm definitely of the opinion that we should wait until we fix the dummy migration script issue. Maybe we should also wait until I or someone else get to try it in their normal work, to spot all the ugly issues tests don't catch.

DillonJ:
> > @manuelma maybe we should be tighter with our words regarding when an MR can be merged or not :-)
> 
> Sure @DillonJ - this time I'm definitely of the opinion that we should wait until we fix the dummy migration script issue. Maybe we should also wait until I or someone else get to try it in their normal work, to spot all the ugly issues tests don't catch.

What is the issue precisely? I included the dummy script in MR... and I used it to successfully update all the example DBs

src/expressions/capacity_margin.jl
```diff
@@ -0,0 +1,182 @@
...
+            # Commodity flows to storage units
+            - sum(
+                unit_flow[u, n, d, s, t_short]
+                for (u, n, d, s, t_short) in unit_flow_indices(
+                    m;
+                    node=n,
+                    direction=direction(:from_node),
+                    stochastic_scenario=s,
+                    t=t,
+                    temporal_block=anything,
```
manuelma, line 88: temporal_block=anything? Does this mean we don't want to use representative_periods for this expression?

src/expressions/capacity_margin.jl
```diff
@@ -0,0 +1,182 @@
...
+                   (unit=u, node=n, stochastic_scenario=s, t=t)
+                   for u in unit__to_node(node=n, direction=direction(:from_node)) if is_storage_unit(u)
+                   for (u, n, d, s, t) in unit_flow_indices(m; unit=u, node=n, direction=direction(:from_node), t=t) 
+                ]
+            ]
+            
+        )
+        
+    )
+end
```
manuelma, line 167: Probably a good idea to test performance in a large model...

src/variables/variable_min_capacity_margin_slack.jl
```diff
@@ -0,0 +1,45 @@
...
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+# GNU Lesser General Public License for more details.
+#
+# You should have received a copy of the GNU Lesser General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#############################################################################
+
+"""
+    node_state_indices(filtering_options...)
+"""
```
manuelma, line 22: Bad docstring.

src/expressions/capacity_margin.jl
```diff
@@ -0,0 +1,182 @@
...
+                    (node=n, stochastic_scenario=s, analysis_time=t0, t=first(representative_time_slice(m, t)))
+                ]
+                * demand[(node=ng, stochastic_scenario=s, analysis_time=t0, t=first(representative_time_slice(m, t)))]
+                for (n, s, t) in node_injection_indices(
+                    m; node=n, stochastic_scenario=s, t=t, temporal_block=anything
+                )
+                for ng in groups(n);
+                init=0,
+            )                                  
+
```
manuelma, line 78: Up until here it looks copy-pasted from another constraint, can we create an expression for the common stuff?

src/expressions/capacity_margin.jl
```diff
@@ -0,0 +1,182 @@
...
+                )
+                if is_storage_unit(u);
+                init=0,
+            )           
+
+            # Conventional and Renewable Capacity
+            + sum(
+                + unit_capacity[(unit=u, node=n_, direction=d, stochastic_scenario=s, analysis_time=t0, t=t)]
+                * unit_availability_factor[(unit=u, stochastic_scenario=s, analysis_time=t0, t=t)]
+                * (
```
manuelma, line 112: No unit_conv_cap_to_flow?

src/data_structure/versions/add_min_capacity_margin.jl
```diff
@@ -0,0 +1 @@
+# Empty migration script - template update is all that is required
```
manuelma, line 1: I think this is a deal-breaker at this point. We need to make it work more robustly because one can get very very lost.

src/data_structure/versions/add_units_out_of_service.jl
```diff
@@ -0,0 +1 @@
+# Empty migration script - template update is all that is required
```
manuelma, line 1: Not sure what this is doing in this PR, but we should avoid it.

src/expressions/capacity_margin.jl
DillonJ: I'm just realising that I don't need to scan the unit_flow variables for active paths since they have to have the same temporal and stochastic structure as node_stochastic_time_indices. So this will be better.

src/expressions/capacity_margin.jl
DillonJ: Only the demand terms are common to `node_injection` everything else is different

src/data_structure/versions/add_min_capacity_margin.jl
DillonJ: @manuelma I must have understood - but what is your idea for handling migrations that only need a template update? A template update will only be triggered if the version increases and the version is the number of upgrade scripts, so what is the alternative? If this can be done quicky, then that's great - otherwise, it's holding up all development?

test/runtests.jl
```diff
@@ -96,14 +101,15 @@ function _dismember_function(func)
...
-@testset begin
+@testset begin    
     include("data_structure/migration.jl")
     include("data_structure/check_data_structure.jl")
     include("data_structure/preprocess_data_structure.jl")
     include("data_structure/temporal_structure.jl")
     include("data_structure/stochastic_structure.jl")
     include("data_structure/algorithm_mga_structure.jl")
     include("data_structure/postprocess_results.jl")
+    include("expressions/expression_tests.jl")
```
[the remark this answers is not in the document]
DillonJ: Not sure - objective and util only have a single file and looking at the top-level files, it doesn't quite fit in

src/data_structure/versions/add_min_capacity_margin.jl
manuelma: Maybe you're right, maybe we can use the empty migration script but two things:

- We don't need a separate file and function per migration. Just write a single funtion in migrations.jl called "_do_nothing" or something like that that returns true, and insert that function as many times as needed and in as many positions as needed within the _upgrade_functions Array.
- It doesn't work if the user runs with upgrade=false and that will be a headache for our users that pull the master until we do something. We can keep developing and helping those users as they come, but I would like to do something rather soon.

test/runtests.jl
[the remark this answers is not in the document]
manuelma: I just don't like the `_tests` part of the file name, we know those are tests and no other file has that even though they are also tests. If you could rename it to `basic.jl` or even `capacity_margin.jl` that would be better.

src/expressions/capacity_margin.jl
manuelma: That's fine, if we want a 'demand' expression we can always do it later using this same mechanism.

src/data_structure/versions/add_min_capacity_margin.jl
manuelma: Actually, run_migrations imports the template at the end regardless of the version. Why did you find that you needed to bump the version in order to load the template?

Did you try running with upgrade=true and without bumping the version in the template?

manuelma:
The only issue is tests failing on windows... will try and fix, then merge.

PR #928, opened 2024-03-08
Attempt to fix the issue #927

### LINKED ISSUE #927: unidirectional connection not working with rolling
When the relationship `connection__from/to_node` for one direction is missing, an error is thrown 
```
BoundsError: attempt to access 1-element Vector{SpineInterface.Object} at index [2]
```
which is caused by [this syntax](https://github.com/spine-tools/SpineOpt.jl/blob/4eb1767eb2a5aacaa067d6c1dec0d823b6fd200d/src/data_structure/postprocess_results.jl#L57) for saving the solved values as moving on.

The issue can be reproduced by running 
[test_in.zip](https://github.com/spine-tools/SpineOpt.jl/files/14528222/test_in.zip)

nnhjy:
In the original code, it is assumed the `connection__from_node` contains at least 2 `node`s, which could not be the case where a connection has only single direction. Unless I misunderstand the function for saving the average through flow, changes are made to fetch the 2nd `node` `n_to` from `connection__to_node`.

Fixes #927 

## Checklist before merging
- [x] Unit tests pass

### DIFF
--- a/src/data_structure/postprocess_results.jl
+++ b/src/data_structure/postprocess_results.jl
@@ -53,8 +53,11 @@ function _save_connection_avg_throughflow!(m::Model, key, connection_flow)
     sizehint!(avg_throughflow, length(connection()) * length(stochastic_scenario()) * length(time_slice(m)))
     for ((conn, n, d, s, t), value) in connection_flow
         start(t) >= m_start || continue
-        # NOTE: always assume that the flow goes from the first to the second node in `connection__from_node`
-        n_from, n_to, _other_nodes... = connection__from_node(connection=conn, direction=anything)
+        # NOTE: always assume that the flow goes from the first node in `connection__from_node` 
+        # to the first node in `connection__to_node` s.t. only one node for each direction is needed
+        # CAUTION: would cause trouble for multi-terminal connections (with >= 3 nodes linked)
+        n_from, n_from_others... = connection__from_node(connection=conn, direction=anything)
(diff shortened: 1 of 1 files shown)


src/data_structure/postprocess_results.jl
```diff
@@ -53,8 +53,11 @@ function _save_connection_avg_throughflow!(m::Model, key, connection_flow)
     sizehint!(avg_throughflow, length(connection()) * length(stochastic_scenario()) * length(time_slice(m)))
     for ((conn, n, d, s, t), value) in connection_flow
         start(t) >= m_start || continue
-        # NOTE: always assume that the flow goes from the first to the second node in `connection__from_node`
-        n_from, n_to, _other_nodes... = connection__from_node(connection=conn, direction=anything)
+        # NOTE: always assume that the flow goes from the first node in `connection__from_node` 
+        # to the first node in `connection__to_node` s.t. only one node for each direction is needed
+        # CAUTION: would cause trouble for multi-terminal connections (with >= 3 nodes linked)
+        n_from, n_from_others... = connection__from_node(connection=conn, direction=anything)
+        n_to, n_to_others... = connection__to_node(connection=conn, direction=anything)
```
manuelma, line 60: Thanks for this @nnhjy - although I think there are many other places in the code where we use this assumption. Changing it here and not everywhere else may lead to trouble. Could you please double check that the entire source code is consistent after this change?

nnhjy: Ah sure, I will check it

PR #933, opened 2024-03-12
Update Julia version in CI workflow

datejada:
Update Julia version in CI workflow to test the latest release of Julia v1.x

Fixes # 932

## Checklist before merging
- [x] Unit tests pass


abelsiqueira:
Since the coverage change is small (0.05%), I would ignore it.
Also, I would add a conditional to only send coverage for a specific Julia version + OS, like [this](https://github.com/TulipaEnergy/TulipaEnergyModel.jl/blob/76c93edc2a3ab3635bd8548fde02184bd50b00af/.github/workflows/TestOnPR.yml#L41).

Issue #855, opened 2024-01-18
Model generation speed


jkiviluo:
SpineOpt is slow in generating large models as has been reported by many users. This is especially problematic for non-rolling models where the model generation is larger part of the total execution time.

These performance issues may be linked to the widespread use of dicts instead of array type parameters. @jd-lara gave us good pointers recently and we'll be pursuing those. There is reason to believe that the situation can be considerably improved.

@Tasqu is doing profiling at the moment and we'll discuss the steps forward in the Mopo project meeting 22-26th Jan. in Leuven. This issue will then hopefully receive sub-issues detailing what to be done next.

DillonJ:
Just to capture here that in conversations with @manuelma he thought that the problems may be in "the joining and unique calls, that require intermediates allocations"

manuelma:
> These performance issues may be linked to the widespread use of dicts instead of array type parameters

Emphasis on **may be** - we do not know yet. @Tasqu maybe you can also profile allocations, see [here](https://docs.julialang.org/en/v1/manual/profile/#allocation-profiler).

My feeling, but of course also to verify, is we do too many allocations. Sometimes we see several GB of allocations for generating a constraint, like 20 GB.

DillonJ:
Also to add here that if we could switch contraints off using a higher level switch like a method parameter - that would also improve model generation speed. Users have resorted to commenting constraints out because the way its currently written, it takes a lot of looping through indexes right now for SpineOpt to know that the constraint indices are a null set. In many cases, we should know that if a parameter has no values, then we don't need to generate certain constraints at all. This is a very low hanging fruit. I will create an issue

manuelma:
> Users have resorted to commenting constraints out because the way its currently written, it takes a lot of looping through indexes right now for SpineOpt to know that the constraint indices are a null set

@DillonJ it would be nice to see/reproduce this experiment, do you have more information?

Tasqu:
@manuelma @DillonJ @jkiviluo If you want, I can spoil the surprise for next week 😄 Or you can spoil yourselves, as I've already uploaded my presentation with the initial results of the profiling I did to the Mopo sharepoint. 

[...]

PR #944, opened 2024-03-21
Performance optimisation

manuelma:
This PR (together with a recent commit to SpineInterface) improves the algorithm to compute the low/highest resolution time slices in a set. Before, for each time slice we would look at all the others to see if it contained/was contained in the other. This has a complexity of order n square, leading to quick performance degration with the number of time slices. Now we rely on ordering/grouping time slices by duration to achieve near order n performance.

I tested Yi's PyPSA model and the 1 year run now completes in about 2 minutes and a half instead of the 10 minutes previously reported.

Re #855 

## Checklist before merging
- [x] Unit tests pass


DillonJ:
Very very nice.... a 5x speedup is not bad at all - nice work!

PR #946, opened 2024-03-22
Configure & activate JuliaFormatter

### LINKED ISSUE #517: Use an automatic code formatter
We discussed using a code formatter to unify our style, @Tasqu mentioned https://github.com/domluna/JuliaFormatter.jl ? It seems like it's the way to go. Any opinions?

clizbe:
(description of changes)

Fixes #517

## Checklist before merging
- [X] Code has been formatted according to SpineOpt's style
- [NA] Unit tests pass


.JuliaFormatter.toml
```diff
@@ -1,6 +1,14 @@
...
+whitespace_in_kwargs = false
+format_docstrings = false
+separate_kwargs_with_semicolon = false
+verbose = true
+whitespace_typedefs = false
+join_lines_based_on_source = true
+surround_whereop_typeparameters = true
+trailing_comma = "nothing"
+align_conditional = true
+annotate_untyped_fields_with_any = true
```
manuelma, line 14: annotate_untyped_fields_with_any = true? I think this should be false.

trailing_comma = "nothing"? we could have true here I think.

clizbe: Yeah we can set untyped_fields to false.
For the trailing_comma, I set it to nothing because it seems we like using it in some cases and not in others, and I assumed there was a reason. But if there's no reason we can just enforce it. :)

PR #947, opened 2024-03-22
Add the objective function of the examples to the tests in run examples.jl

### LINKED ISSUE #935: Improve the `run_examples.jl` tests to include the objective function
Testing the objective function of the current examples is a way to have some integration tests that at least cover the functionalities in the tutorials.

datejada:
To maintain consistency in the objective function of our examples, we add it to the test. We use a dictionary to select the examples we want to test the objective function for, or simply check if it's optimal.

Fixes #935 

## Checklist before merging
- [x] Unit tests pass

### DIFF
--- a/examples/reserves.json
+++ b/examples/reserves.json
@@ -1098,7 +1098,7 @@
                             "data": [
                                 [
                                     "mip_rel_gap",
-                                    0.01
+                                    1e-6
                                 ],
                                 [
                                     "presolve",
--- a/examples/simple_system.json
+++ b/examples/simple_system.json
@@ -1076,7 +1076,7 @@
                             "data": [
                                 [
                                     "mip_rel_gap",
-                                    0.01
+                                    1e-6
                                 ],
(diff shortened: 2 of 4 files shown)


datejada:
@DillonJ or @tarskul, this is the first step in creating more integrated tests in SpineOpt. Please let me know if you have any comments before merging into the master. Thanks!

tarskul:
To me it seems too strict at the moment. If the examples were a bit more theoretical  where we know the results with pinpoint accuracy (e.g. 100.0) then it would make sense to be this strict. But these are a bit more practical so we can expect some rounding errors etc. So I would allow for a small margin before we get alerted that a unit test fails.

datejada:
@tarskul Thanks! I agree that the 1e-6 tolerance for tests is too much. Even with the solver tolerance within that value, increasing it to avoid false negatives due to rounding errors is better. Let's use 0.1% for now and start testing to see how it goes.

PR #952, opened 2024-03-26
Upate cis file for new user meeting

clizbe:
Changed file to new meeting from Tars. Wasn't sure where to save it so just put it in general SpineOpt.jl directory. Hope that works - can't really check without uploading the file.

Fixes # (issue)

## Checklist before merging
- [NA] Unit tests pass


tarskul:
As you said, it is a bit of an awkward file to have floating around. If it is in the way of the developers it can move to the project level folder .github. But for now it is fine.

PR #957, opened 2024-03-28
Change test `run_examples.jl`

### LINKED ISSUE #956: Change `run_examples.jl` so that tolerances for mip cases are higher
In the example jsons, parameters `mip_rel_gap` was changed from 0.01 to 1e-6 after this commit https://github.com/spine-tools/SpineOpt.jl/commit/c0b4a386f9806fbd77d0471db830b325a1f7db9f.

However, the default value in the template is still 0.01. This means after upgrading the jsons using `upgrade_all.jl`, it will be set back to 0.01. Having this gap will fail the integration test at `run_examples.jl` because the test was written with a gap of 1e-6.

We only want to strictly test the LP cases, but not the MIP cases.

gnawin:
I did two things:
- change the test
- run `upgrade_all.jl`; that's why there are more changes in the jsons.

Fixes #956 

## Checklist before merging
- [x] Unit tests pass

### DIFF
--- a/examples/reserves.json
+++ b/examples/reserves.json
@@ -356,7 +356,7 @@
                 "connection",
                 "node"
             ],
-            "Defines the `nodes` the `connection` can take input from, and holds most `connection_flow` variable specific parameters.",
+            "A flow on a `connection` from a `node`.",
             280378317271897
         ],
         [
@@ -366,7 +366,7 @@
                 "node",
                 "investment_group"
             ],
-            "Indicates which connection capacities are included in the capacity invested available of an investment group",
+            "A flow on a `connection` from a `node` whose capacity should be counted in the capacity invested available of an `investment_group`.",
             null
         ],
         [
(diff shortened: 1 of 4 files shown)


datejada (approved):
LGTM. Thanks!

gnawin:
The codecov fails and reports 2 lines are missing from `src/run_spineopt.jl`, these are line 169 and 170. As far as I understand, these two lines check if anything is missing from the input database.  In this PR, I updated the json files according to the latest template, so there are no missing items. 
Is it okay if I merge this PR? @manuelma @Tasqu 

```
166 | 166 | template, data = _init_data_from_db(url_in, log_level, upgrade, templates, filters)1
167 | 167 | missing_items = difference(template, data)1
168 | 168 | if !isempty(missing_items)1
  169 | 169 | println()
  170 | 170 | @warn """
  171 | 171 | Some items are missing from the input database.
  172 | 172 | We'll assume sensitive defaults for any missing parameter definitions,
  173 | 173 | and empty collections for any missing classes.

```

PR #970, opened 2024-04-05
Add free energy warning to docs

### LINKED ISSUE #414: Add warning in documentation about history flows
When one uses the `connection_flow_delay` parameter the initial flows on the output side of the relevant connection are related to flows on the input side for timeslices before the model window, which can be unconstrained which can result in free energy for the model.

In my case with storage investments, the model was using this to fill up an energy store for free.

A work around presently is to use `fix_connection_flow` to fix the flow to zero for the history timeslices. 

Regarding an enduring solution... I can only really think of one option and that is to fix the history flows to zero automatically in  pre-processing but I feel this isn't ideal as perhaps it would result in infeasibilities in come cases. Another option may be to warn the user about this potential problem in the documentation.

clizbe:
Fixes #414

## Checklist before merging
- [X] Code has been formatted according to SpineOpt's style
- [NA] Unit tests pass


clizbe:
@DillonJ Can you please review this PR so I can merge?

Issue #909, opened 2024-03-01
Create new parameters for the economic representation in the multi-year investments


datejada:
Taking as a reference the changes in the file `economic_structure` in PR #448 we need:

- [x] List the new parameters and relationships with explanations in this issue
- [x] Create the new parameters and relationships with their defaults in the SpineOpt template
- [x] Create a new case study in JSON format from the template with data (maybe future use as the tutorial)
- [x] Create a new file with the economic representation and add the function from #448 
- [x] Refactor the code as suggested in #448
- [x] Create the new tests for the economic representation
- [ ] Create migration script for renamed parameters

Notice that with this issue, we just create the parameters and calculations but we don't use them yet in the model

manuelma:
Here it would be very useful to have a list of the parameters and their meaning.

gnawin:
@manuelma most of the information is indeed in #476 but they can improved for sure. I can start by listing and explain things here, and later we see where the information is best placed. Note that they may be repetitive to what is there already in #476, but I agree at the moment it is important to explain more without worrying too much about being repetitive.

manuelma:
Thank you @gnawin - I need to go through https://github.com/spine-tools/SpineOpt.jl/discussions/476 indeed.

gnawin:
Below is a list of parameters for current design and the new design. More explanation of the parameters (including a document with more details) is in discussion #476.


# Current design

## Current parameters
`connection_investment_lifetime`, `storage_investment_lifetime`, `unit_investment_lifetime`


# New design

## Modified parameters
`connection_investment_tech_lifetime`, `storage_investment_tech_lifetime`, `unit_investment_tech_lifetime`


## New parameters

### User-defined parameters

- discount: `discount_rate`, `discount_year`

- lead time: `unit_lead_time`, `connection_lead_time`, `storage_lead_time`

- life time: `connection_investment_economic_lifetime`, `storage_investment_economic_lifetime`, `unit_investment_economic_lifetime`

### Internally calculated parameters

- annuities: `unit_conversion_to_discounted_annuities`, `connection_conversion_to_discounted_annuities`, `storage_conversion_to_discounted_annuities`
- capacity transfer factor: `unit_capacity_transfer_factor`, `connection_capacity_transfer_factor`, `storage_capacity_transfer_factor`
- salvage: `unit_salvage_fraction`, `connection_salvage_fraction`, `storage_salvage_fraction`
[...]

PR #973, opened 2024-04-10
Iterators

manuelma:
Some optimisations:
- Use iterators for constraint and variable indices rather than vectors.
- Don't traverse iterators given to `active_stochastic_paths()` unless there's more than one `stochastic_scenario` in the model.
- Remove `units_available` and keep only `units_on`.
- Don't build the auto-updating structure if the model is not rolling.
- Don't generate unit commitment variables unless they are needed.
- Use variable bounds instead of constraints whenever possible.
- `use_connection_intact_flow` is false by default.
- Introduce `use_tight_compact_formulations`, false by default.

Re https://github.com/spine-tools/SpineOpt.jl/issues/855

## Checklist before merging
- [ ] Unit tests pass


gnawin:
@manuelma, I've experienced the same errors for Julia 1.6 and only for 1.6 in PR #909. Do you happen to know how to fix it? Thanks.

manuelma:
I am splitting this into more self-contained PRs - one for each optimisation. Closing now.

PR #975, opened 2024-04-11
Apply lookahead relative to the end of the window

manuelma:
In Benders, if user defines block_end it is safe to assume it is relative to the sub problem rolling window. Currently, if the user defines a look-ahead for the investment block so it is applied to the rolling window, it is also applied to the master problem window with disastrous consequences (the investment block will end too soon in the master problem, probably).

This PR fixes the above issue by adjusting the block end in the master problem window so it looks ahead the same amount as in the subproblem rolling window.

## Checklist before merging
- [x] Unit tests pass


manuelma:
Trying to explain this in the code I realized it's not a very good idea.

PR #981, opened 2024-04-18
Don't generate auto-update structure if not rolling and no benders

manuelma:
If no rolling and no benders we don't need to build the auto-update structure which will save us some work.

## Checklist before merging
- [x] Unit tests pass


manuelma:
The strategy is simple: instead of calling directly `getindex` on our `Parameter`s (i.e., `parameter_name[(kwargs...)]`, the 'brackets' call), we use an intermediate function that receives the model as first argument, and then dispatches to `getindex` or just calls the `Parameter` depending on whether the model is rolling or benders, or not.

PR #1013, opened 2024-05-12
Missing uc cases

nnhjy:
add some more cases of unit commitment where variable `units_on` is needed 

Fixes # (issue)

## Checklist before merging
- [x] Unit tests pass

### DIFF
--- a/test/run_spineopt_investments.jl
+++ b/test/run_spineopt_investments.jl
@@ -95,6 +95,7 @@ function _test_capacity_investments()
         url_in, url_out, file_path_out = _test_run_spineopt_investments_setup()
         object_parameter_values = [
             ["model", "instance", "use_connection_intact_flow", false],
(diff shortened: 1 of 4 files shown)


test/run_spineopt_investments.jl
```diff
@@ -95,6 +95,7 @@ function _test_capacity_investments()
         url_in, url_out, file_path_out = _test_run_spineopt_investments_setup()
         object_parameter_values = [
             ["model", "instance", "use_connection_intact_flow", false],
+            ["unit", "unit_a", "units_on_cost", 0], # Just to have units_on variables
```
manuelma, line 98: Nice catch @nnhjy !

In my opinion this is good to merge, the only question is why this. Why do we need it?

nnhjy: Ah, I overlooked the case where the investment mode requires `units_on` to bound investment transition. I will fix this. Thanks for pointing this out.

PR #1036, opened 2024-06-24
Changed the default number of units based on is_candidate.

Mastomaki:
This one adds the same functionality for units as is already for storages and connections, namely overriding the default number of units.

However, it can be confusing to the user because as far as I see the overriding has not been mentioned in the documentation.


DillonJ:
Nice work @Mastomaki - this makes a lot of sense - looks like all tests are passing and it can be merged. @manuelma does this look ok to you?

src/constraints/constraint_units_available.jl
```diff
@@ -53,10 +53,15 @@ function _build_constraint_units_available(m, u, s, t)
...
         <=
-        number_of_units(m; unit=u, stochastic_scenario=s, analysis_time=t0, t=t)
+        number_of_units(m; unit=u, stochastic_scenario=s, analysis_time=t0, t=t,
+            _default=_default_number_of_units(u))
     )
 end
 
+# Change the default number of units so that it is zero when candidate units are present
+# and otherwise 1.
+_default_number_of_units(u) = is_candidate(unit=u) ? 0 : 1
```
manuelma, line 63: Should the `1` be `number_of_units(...)` instead?

manuelma: Also, unit tests?

Mastomaki: I now changed the code to reflect this and updated the unit test.

src/constraints/constraint_units_available.jl
```diff
@@ -53,7 +53,9 @@ function _build_constraint_units_available(m, u, s, t)
             init=0,
         )
         <=
-        + number_of_units(m; unit=u, stochastic_scenario=s, t=t)
+        # Change the default number of units so that it is zero when candidate units are present
+        # and otherwise 1.
+        + ifelse(is_candidate(unit=u), 0, number_of_units(m; unit=u, stochastic_scenario=s, t=t) )
```
manuelma, line 58: Doesn't this disallow having both number_of_units and candidate_units different than zero? It looks like if the user specifies candidate_units different that zero then any non zero value for number_of_units is just ignored? Or I am missing something?

Mastomaki: Indeed. Now I tried to correct it.

PR #1058, opened 2024-08-09
Include discounted duration parameter to outputs of the economic representation

### LINKED ISSUE #1057: Include unit_discounted_duration, connection_discounted_duration, storage_discounted_duration in the output

datejada:
(description of changes)

Fixes #1057 

## Checklist before merging
- [x] Unit tests pass

### DIFF
--- a/src/run_spineopt_basic.jl
+++ b/src/run_spineopt_basic.jl
@@ -671,6 +671,7 @@ function _save_economic_parameter_values!(m::Model)
     m.ext[:spineopt].values[:unit_salvage_fraction] = Dict{key_type,Float64}()
     m.ext[:spineopt].values[:unit_tech_discount_factor] = Dict{key_type,Float64}()
     m.ext[:spineopt].values[:unit_conversion_to_discounted_annuities] = Dict{key_type,Float64}()
+    m.ext[:spineopt].values[:unit_discounted_duration] = Dict{key_type,Float64}()
     if use_economic_representation(model=m.ext[:spineopt].instance)
         _populate_economic_parameter_values!(
             m,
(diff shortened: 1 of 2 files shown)


src/run_spineopt_basic.jl
```diff
@@ -639,25 +639,15 @@ end
...
-        Tuple{SpineInterface.Object, SpineInterface.Object, SpineInterface.TimeSlice}
-    }
-    
-    if !use_economic_representation(model=m.ext[:spineopt].instance)
-        m.ext[:spineopt].values[param_name] = Dict{key_type,Float64}()
-        return nothing
-    end
-
     m.ext[:spineopt].values[param_name] = 
         Dict{key_type,Float64}(
```
manuelma, line 652: Why do you need to specify the key_type of the dictionary? I believe just `Dict(...)` will have the same effect here - remember Julia infers the type from the element when possible and I believe this is one of these cases.

datejada: Thanks, @manuelma; I agree with your comment. I did it because the [Style Guidelines](https://github.com/spine-tools/SpineOpt.jl/blob/master/CONTRIBUTING.md#style-guidelines) proposes it: _"Never leave things without type qualifications. Use an explicit ::Any."_

Unless I didn't understand the Style Guideline (which is a possibility 😝).

Anyway, I think being explicit about types is better for new developers when reading the code and does not harm performance.

manuelma: Do you really find it easier to read? If that's the case then nevermind ;)

manuelma: Btw it looks like we copy-pasted those guidelines from somewhere else - also the previous point "Don't explicitly parameterize types unless it's necessary" seems to contradict the one you're referring?

datejada: Okay! Maybe we just need to update the guidelines, and then we can simplify this, as you suggest. What do you think?

PR #1060, opened 2024-08-13
Update technical lifetime constraints to include sense

### LINKED ISSUE #689: Update on the technical lifetime of a unit/storage/connection constraint
Create a method so that the user can select either ">=" or "==" in the technical lifetime constraint and update the documentation

https://spine-tools.github.io/SpineOpt.jl/latest/mathematical_formulation/constraints/#Technical-lifetime-of-a-unit

datejada:
This change includes a new parameter `_investment_lifetime_sense` for units, connections, and storages. The parameter defines the sense of the lifetime constraint. The options are ">=", "==", and "<=", to give the user to consider minimum lifetime, strict lifetime, or maximum lifetime. By default, the parameter is strict, i.e., "==", according to the review of the formulation in the Mopo project.

Fixes #689 

## Checklist before merging
- [x] Unit tests pass


tarskul:
Seems great work to me!

I'm only wondering about the change of the default behaviour. Yes, it may be better in general. But I'm thinking about existing projects. Will some of them suddenly break without them understanding why?

We should make a migration script where existing databases set the parameter explicitly to the inequality.

We can also retain the inequality for now. Then we merge and still have some time to make a migration script. Once we have the migration script, we can then make the change in default behaviour.

At the very least we should communicate it clearly.

DillonJ:
I agree with @tarskul that the default behaviour should be the existing behaviour so that existing models are not impacted.

datejada:
> I agree with @tarskul that the default behaviour should be the existing behaviour so that existing models are not impacted.

Ok! It makes sense. I can make that change before merging this PR. Thanks!

test/constraints/constraint_node.jl
```diff
@@ -845,6 +845,67 @@ function test_constraint_storage_lifetime()
...
+            var_s_inv_av_key = (node(:node_c), s, t)
+            var_s_inv_av = var_storages_invested_available[var_s_inv_av_key...]
+            vars_s_inv = [var_storages_invested[node(:node_c), s, t] for (s, t) in zip(s_set, t_set)]
+            expected_con = SpineOpt.build_sense_constraint(var_s_inv_av - sum(vars_s_inv), sense_value, 0)
+            observed_con = constraint_object(constraint[key...])
+            @test _is_constraint_equal(observed_con, expected_con)
+        end
+    end
+end
+
```
manuelma, line 908: Am I correct that this test is just a copy paste of `test_constraint_storage_lifetime()` but adding the sense? In this case we could get rid of `test_constraint_storage_lifetime()` because all it is testing is now tested in the new test. What do you think @datejada ?

datejada: Hi @manuelma, thanks! The test `test_constraint_connection_lifetime()` is for 3 different lifetime values, whereas the `test_constraint_connection_lifetime_sense` test the 3 different senses for one specific lifetime. I agree that we could join both tests in one, but I am not sure why because I want to avoid creating 3x3 runs in the test and only have 3 + 3. Also, it seems that the difference in the number of runs is not relevant, but I want to avoid doing extra tests. I am open to suggestions or changes.

PR #1074, opened 2024-08-28
Calculate the `_discounted_duration` outputs with the flow indices

### LINKED ISSUE #1072: Fix `X_discounted_duration` output to include the operational
**Describe the bug**

The current version of the code only prints for the investment years. The `X_discounted_duration` should be for the operational years.

datejada:
The output for the `_discounted_duration` should now use the flow indices instead of the investment ones. 

Fixes #1072 

## Checklist before merging
- [x] Unit tests pass

### DIFF
--- a/src/run_spineopt_basic.jl
+++ b/src/run_spineopt_basic.jl
@@ -633,9 +633,9 @@ function _populate_economic_parameter_values!(
 )
     m.ext[:spineopt].values[param_name] = 
         Dict{key_type,Float64}(
-            (entity=>e, stochastic_scenario=s, t=t) =>
-            value_func(; entity=>e, stochastic_scenario=s, t=t)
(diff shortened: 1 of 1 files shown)


datejada:
@gnawin, would you mind testing these changes in your case study? (you can switch to this branch since it is in the main repo)

The output for the `_discounted_duration` should now use the flow indices instead of the investment ones. 

Note: It is only for `units`  and `connections` since `storage` is not used in the current state of the model.

gnawin (approved):
@datejada Thanks. I checked the parameter `unit_discounted_duration`, and indeed it correctly shows the parameters at the resolution of the unit_flow_indices. I did not test connection, because I don't have a connection in my test case, but it should work the same way. Feel free to merge if you agree.

PR #1076, opened 2024-08-28
Piecewiselinear efficiency using ratio unit flow

### LINKED ISSUE #444: Piecewiselinear efficiency using fix_ratio_out_in_unit_flow?
I can set up the piecewiselinear efficiency using the incremental_heat_rate parameter, however I cannot do it using fix_ratio_out_in_unit_flow. Is this intentional? I thought this feature 'duplication' was to please different worldviews, but one is more powerful than the other?

nhniina:
Incremental and idle heat rates have been removed and their functionality has been replaced by allowing array type `fix_ratio_in_out_unit_flow` together with `fix_units_on_coefficient_in_out` and `operating_points`.

Note: Tests cover a case when the user gives array type `fix_ratio_in_out_unit_flow`. However, they do not cover a case when a user tries to give array type `fix_ratio_out_in_unit_flow`, `min_ratio_in_out_unit_flow`, etc.

Fixes #444

## Checklist before merging
- [x] Unit tests pass


tarskul:
Seems good work to me. I've gone through it at a steady pace so I might have missed something but we can always fix it afterwards.

I was a bit confused about some of the changes in the examples but I guess that is because they probably weren't updated in a while. But the tests pass so it is probably ok. It does seem that the euro symbol may be broken (at least github does not seem to render the in in the updated files). Maybe we need to change those to EUR in the template as to avoid this issue? But it doesn't seem too important.

DillonJ:
Nice work @nhniina. I've added @manuelma as a reviewer. We need to make sure this plays nice with what he's done with the replacement expressions. They have a big payoff in terms of model size, but at the expense of quite some complexity and we have already had a few issues relating to it. I think we might need some more tests that test the functionality at a higher level. For example units with multiple outputs and units with multiple inputs (this is where we had issues with the replacement expressions before).

Thoughts @manuelma

src/constraints/constraint_ratio_unit_flow.jl
```diff
@@ -223,7 +264,9 @@ function constraint_ratio_unit_flow_indices(m::Model, ratio)
     (
         (unit=u, node1=n1, node2=n2, stochastic_path=path, t=t)
         for (u, n1, n2) in indices(ratio)
-        if !_has_simple_fix_ratio_unit_flow(n1, n2, ratio)
+        if !_has_simple_fix_ratio_unit_flow(n1, n2, ratio) || 
+            !isempty(unit_flow_op_indices(m; unit=u, node=n1, direction=d1)) || 
+            !isempty(unit_flow_op_indices(m; unit=u, node=n2, direction=d2))
```
manuelma, line 269: I think we should put the two new checks inside the `_has_simple_fix_ratio_unit_flow` function for readability.

src/variables/variable_unit_flow.jl
```diff
@@ -139,7 +139,9 @@ function add_variable_unit_flow!(m::Model)
             ),
         )
         for (u, n_ref, d_ref, n, d, fix_ratio, direct) in _related_flows(fix_ratio_d1_d2)
-        if _has_simple_fix_ratio_unit_flow(n, n_ref, fix_ratio)
+        if _has_simple_fix_ratio_unit_flow(n, n_ref, fix_ratio) && 
+            isempty(unit_flow_op_indices(m; unit=u, node=n_ref, direction=d_ref)) && 
+            isempty(unit_flow_op_indices(m; unit=u, node=n, direction=d))
```
manuelma, line 144: Again, the two new checks should go inside `_has_simple_fix_ratio_unit_flow` so we don't need to duplicate the code here and in the fix_ratio constraint.

src/data_structure/versions/translate_heatrate_parameters.jl
```diff
@@ -0,0 +1,99 @@
...
+	end
+	pdef = run_request(db_url, "call_method", ("get_parameter_definition_item",), Dict(
+		"entity_class_name" => "unit__node__node", "name" => "unit_incremental_heat_rate")
+	)
+	if length(pdef) > 0
+		run_request(db_url, "call_method", ("remove_parameter_definition_item", pdef["id"]))
+	end
+
+	true
+end
```
manuelma, line 99: This migration function uses 0.8 spinedb_api so we cannot merge these changes into SpineOpt's 0.7 branch, directly - we'd need to write a new migration function using the corresponding 0.7 spinedb_api.

manuelma: I also just remembered there was some discussion about keep supporting the heat rate parameters but introduce a new prepocessing step to translate them into fix-ratio parameters. That would remove the need for the migration script but of course the question is do we want to keep supporting the heat rate parameters?

src/constraints/constraint_ratio_unit_flow.jl
nhniina: That's what I would have liked to have as well, but m, unit and direction were not given to `_has_simple_fix_ratio_unit_flow` so I didn't want to start changing that function too much. But if you think that's the cleaner way, let's try to update code accordingly.

manuelma: Hi @nhniina - yes I think so, because we also do the checks in some other place (when creating the variable) so it's better if we don't repeat that code. Thanks! Great work btw :)

manuelma:
Should I press the button @nhniina ?

nhniina:
Yes, please, if you think the codecov 1 line is not a problem.

PR #1082, opened 2024-09-03
Add multi-year information to advanced concepts in the doc

### LINKED ISSUE #913: Update the documentation (e.g., formulation and features)
For multi-year investment

gnawin:
(description of changes)

Fixes #913 

## Checklist before merging
- [ ] Unit tests pass


gnawin:
@datejada I ask for your review because we discussed what to include in this doc in the past, you could double-check to see if any improvements are needed. Also this is the last issue of the epic. Thank you!

datejada (approved):
Thanks for the PR. I am unsure of storing PDF and Excel files in the repo. But there is no better way for now. The good thing is that they are small files. So, Let's ask during the developers' meeting if they have a better idea or if we just continue like this for the time being.

manuelma:
> Hi @manuelma I also want to add a warning, saying that we don't support rolling/MGA/Benders in combination with economic representation. For rolling, I know I can check the parameter `roll_forward`, but what about MGA/Benders, are there similar parameters that I can use as conditions? Thank you.

Why don't we support it?

gnawin:
> > Hi @manuelma I also want to add a warning, saying that we don't support rolling/MGA/Benders in combination with economic representation. For rolling, I know I can check the parameter `roll_forward`, but what about MGA/Benders, are there similar parameters that I can use as conditions? Thank you.
> 
> Why don't we support it?

We are doing changes in run_spineopt_basic.jl, and I reckon Benders is a different file. For MGA I'm not exactly sure. My understanding is they would require some "modifications" to the usual temporal and/or stochastic structures. These specifications were not straightforward to see/do for the new parameters as the focus was to start with the basic case. So to be on the safe side, I'm inclined to add a warning. It can also be that I'm wrong due to ignorance. If you could have a better judgement there, that would be great!

manuelma:
> > > Hi @manuelma I also want to add a warning, saying that we don't support rolling/MGA/Benders in combination with economic representation. For rolling, I know I can check the parameter `roll_forward`, but what about MGA/Benders, are there similar parameters that I can use as conditions? Thank you.
> > 
> > 
> > Why don't we support it?
> 
> We are doing changes in run_spineopt_basic.jl, and I reckon Benders is a different file. For MGA I'm not exactly sure. My understanding is they would require some "modifications" to the usual temporal and/or stochastic structures. These specifications were not straightforward to see/do for the new parameters as the focus was to start with the basic case. So to be on the safe side, I'm inclined to add a warning. It can also be that I'm wrong due to ignorance. If you could have a better judgement there, that would be great!

Correct me if I'm wrong, but all we're doing regarding the 'economic representation' is take some 'source' parameter values defined by the user in the DB and translating them into some other (more useful) parameter values to be used in the model, based on some excel-level calculations.

I don't see why these new parameter values couldn't be used with rolling, benders or mga, is there anything special to them that makes them incompatible?

Also, I was thinking on what we said in the last meeting, that generating these economic parameters requires the temporal and stochastic structures, but why is that? Couldn't the transformation be based exclusively on the data provided for the 'source' parameters?

gnawin:
@manuelma Sorry for the late reply, I wanted to go through the code and provide some details so it required a bit of time. Maybe I can briefly explain how the parameters are calculated, then we can together see if there are ways to improve.

The main point is that the calculation itself is indeed Excel level, which is based on some temporal and stochastic structures. 

To calculate an investment-related parameter, we would need the investment indices and the associated stochastic structure. 

For the operation-related parameter `discounted_duration`, there are two ways of calculating this: `use_milestone_years=true` or `use_milestone_years=false`. For the former, it's using the same structures as the investment ones. And for the latter, we just use all the model years (i.e., counting from the first model year to the last model year), meaning two things 1) since it just depends on the model horizon, does it work with mga/rolling/Benders? 2) we don't incorporate stochastic structure either there. The latter is less developed because of the formulation itself and the fact that this will be less used. 

Based on this, do you have any suggestions about what we support or not, and in what cases/algorithms?

test/data_structure/check_economic_structure.jl
```diff
@@ -273,4 +273,52 @@
...
+            ["unit", "unit_ab", "candidate_units", candidate_unts],
+            ["unit", "unit_ab", "unit_investment_cost", inv_cost],
+            ["unit", "unit_ab", "unit_discount_rate_technology_specific", tech_discnt_rate],
+            ["unit", "unit_ab", "unit_lead_time", Dict("type" => "duration", "data" => "1Y")],
+            ["unit", "unit_ab", "unit_investment_tech_lifetime", Dict("type" => "duration", "data" => "5Y")],
+            ["unit", "unit_ab", "unit_investment_econ_lifetime", Dict("type" => "duration", "data" => "5Y")],
+        ]
+        SpineInterface.import_data(url_in; object_parameter_values=object_parameter_values)
+        @test_throws ErrorException run_spineopt(url_in; optimize=false, log_level=1)
+    end
```
gnawin, line 299: Here is the test for the line that Codecov is failing to recognize.

gnawin:
Codecov is not passing, but the missing line is tested in https://github.com/spine-tools/SpineOpt.jl/pull/1082/files#r1768334315

PR #1085, opened 2024-09-03
Implement changes from code review

### LINKED ISSUE #1083: Remove specify the key_type of the dictionary for economic parameters outputs
Why do you need to specify the key_type of the dictionary? I believe just `Dict(...)` will have the same effect here - remember Julia infers the type from the element when possible and I believe this is one of these cases.

_Originally posted by @manuelma in https://github.com/spine-tools/SpineOpt.jl/pull/1058#discussion_r1741685969_

datejada:
Include changes on the output of the economic parameters according to code review from @manuelma 

Fixes #1083 

## Checklist before merging
- [x] Unit tests pass


datejada:
@manuelma, would you mind reviewing these changes? They come from the reviews you made early this week to the code. Thanks!

manuelma:
Looks better @datejada 

I am wondering, the purpose of this code is to be able to save the economic parameters as outputs in the output DB, but why this need a special function? Because we are able to save any input parameter to the output DB - shouldn't this use that same path?

datejada:
> Looks better @datejada
> 
> I am wondering, the purpose of this code is to be able to save the economic parameters as outputs in the output DB, but why this need a special function? Because we are able to save any input parameter to the output DB - shouldn't this use that same path?

Hi @manuelma, thanks for your reply 😄

To be honest, I didn't know about that function in the code. I don't know if it works for the economic parameters. I am a bit unsure that it will work because the economic parameters are only created if the economic representation is activated.

I will check and let you know 😉

datejada:
> > param = parameter(out_name, @**MODULE**)
> 
> Thanks @datejada - could you perhaps try replacing `param = parameter(out_name, @__MODULE__)` with just `param = parameter(out_name)` and see if that helps (in line 989)?

🤣 it works, you are the best! 😄 This means that all this code can be scraped if we just change line 989 🤷 Is that correct?

I will run the complete set of tests to be sure it does not fail in another part of the code.

manuelma:
> > > param = parameter(out_name, @**MODULE**)
> > 
> > 
> > Thanks @datejada - could you perhaps try replacing `param = parameter(out_name, @__MODULE__)` with just `param = parameter(out_name)` and see if that helps (in line 989)?
> 
> 🤣 it works, you are the best! 😄 This means that all this code can be scraped if we just change line 989 🤷 Is that correct?
> 
> I will run the complete set of tests to be sure it does not fail in another part of the code.

Thank you @datejada - unfortunately we can't remove the `@__MODULE__` here because that means all `Parameter`s in the SpineOpt module won't be found. The thing is, we attach the economic `Parameter`s to the SpineInterface module (the default), whereas the normal DB input `Parameter`s are attached to the SpineOpt module.

What we could do is attach the economic `Parameter`s to the SpineOpt module too, we can do it by passing the keyword argument `mod` to the `Parameter` constructor.

datejada:
@manuelma ok! thanks. I will try.

BTW: sorry about the master; I was trying the @__MODULE__ thing and didn't realize that I was on master 😢 I reverted the commit.

However, we should avoid committing directly to the master. It is a good practice. What do you think?

datejada:
@manuelma, it works! 😄 Attaching the module to the economic parameter creation will guarantee that the output is created without adding more code. So, I deleted the functions and added the keyword argument to pass the module to the parameter creation.

Let's wait until the tests pass, but it looks nicer now.

Thanks!

datejada:
@manuelma, all checks pass, so your suggestion works. Thanks again. If you are okay with the latest changes, I will merge the PR. Is that OK?

manuelma:
> @manuelma, all checks pass, so your suggestion works. Thanks again. If you are okay with the latest changes, I will merge the PR. Is that OK?

Nice work @datejada - I just merged myself :)

PR #1092, opened 2024-09-11
Add constraint to avoid simultaneous start-ups and shutdowns

### LINKED ISSUE #960: Add constraint to avoid simultaneous starts/shutdowns
Currently, there is nothing preventing simultaneous starts and shutdowns and if there are no start costs, this is seen. It's also used by the model to avoid the ramp_up/down_limit if start_up/shut_down_ramp_limit are not specified.

Are we missing a constraint along the lines of units_started_up + units_shut_down <= units_available?

@datejada @jkiviluo?

datejada:
Add a new constraint to avoid simultaneous start-ups and shutdowns

Fixes #960 

## Checklist before merging
- [x] Unit tests pass


datejada:
@manuelma, this is the PR with the changes we commented on yesterday in the developers' workshop. I would greatly appreciate your review. The main thing I am unsure where to allocate is the warning message we commented on yesterday that it would be nice to have. Do you have any ideas?

@jkiviluo, your feedback would also be appreciated if you have time or are interested.

Thanks both!

datejada:
@manuelma tests passed now 😉

datejada:
After discussion with @g-moralesespana we should add this constraint. The minimum up/down time should be set by default to 1 always to ensure non-simultaneous start-ups and shutdowns (even in the clustered unit commitment case).

So, closing this PR 😞

jkiviluo:
Did you mean 'should **not** add this constraint'? 

I thought having this constraint would anyway make the solution space tighter and would be beneficial - whereas minimum up/down time constraint can be computationally heavy (although not sure how heavy it its for just 1 timestep). So, I thought that using this in case of binary unit commitment should be computationally cheaper. But it seems German has better knowledge.

datejada:
> Did you mean 'should **not** add this constraint'?
> 
> I thought having this constraint would anyway make the solution space tighter and would be beneficial - whereas minimum up/down time constraint can be computationally heavy (although not sure how heavy it its for just 1 timestep). So, I thought that using this in case of binary unit commitment should be computationally cheaper. But it seems German has better knowledge.

Correct, it is 'should **not** add this constraint'. @g-moralesespana argues that for 1 timestep, it is better for the MIP to have the constraints from the minimum up/down time than this constraint that the one here. In any case the minimum up/down should not heavy to compute for 1 timestep (maybe we should check that for SpineOpt).

DillonJ:
Don't forget @datejada that model's aren't always 1h resolution. Does this work when the temporal resolution is greater than 1h?

DillonJ:
We probably can't make any safe assumption about what the default min up/down time should be - unless we process the temporal structure and do something like set it to the duration of the highest resolution timeslice which might be different for each unit

DillonJ:
We probably can't make any safe assumption about what the default min up/down time should be - unless we process the temporal structure and do something like set it to the duration of the highest resolution timeslice which might be different for each unit

g-moralesespana:
The solution should work for whatever resolution. Let's see the  UC case, but the same applies for the clustered UC including investments.
The following constraint is the classic one:
unit_started_up + unit_shut_down <= 1   (1)
Which indeed avoids the simultaneous startup and shutdown once the solution is integer. But to find the integer solution, many LP relaxation problems are solved (during the branch and bound), so ideally, we want to avoid simultaneous charging and discharging even when the variables are relaxed. See how (1) allows unit_started_up = unit_shut_down = 0.5.

If we use the tight formulation of the minimum up/down constraints with the minimum up/down time equal to one (could be longer than an hour), we have:
unit_started_up <= unit_on    (2)
unit_shut_down <= 1 - unit_on   (3)
Notice that if unit_on=1, then (3) forces unit_shutdown=0. And when unit_on=0, then (2) forces unit_started_up=0. So for these situations it is already not possible to have simultaneous charging and discharging. 
Furthermore, you can even define unit_started_up and unit_shut_down as continuous variables, because once unit_on is defined binary, (2) and (3) force unit_started_up and unit_shut_down to only take binary values (together with the commitment logic constraint: unit_on(t) - unit_on(t-1) = unit_started_up(t) - unit_shut_down(t) ).
So (2) and (3) are way tighter than (1), so it is expected to speedup MIP solving times. And even if you solve the problem as a relaxed LP, (2) and (3) have a more accurate approximation, the the formulation is tighter, i.e., the relaxed solutions are nearer to the integer one.

p.s., Although unit_started_up and unit_shut_down can be defined continuous thanks to (2) and (3), it is better, and expected to be computationally more efficient if they are defined as integer.

datejada:
> Don't forget @datejada that models aren't always 1h resolution. Does this work when the temporal resolution is greater than 1h?
> 
> 

Correct, that is why the comment says ‘1 timestep’ instead of ‘1h’. The explanation from German shows why we say it like that.

PR #1121, opened 2024-10-24
1116 add dual variables into the output list in the template

### LINKED ISSUE #1116: Add dual variables into the output list in the template

nnhjy:
1. Add in the current template the dual values as output 
2. Add a dummy parameter to classify the output values

Fixes #1116 

## Checklist before merging
- [x] Unit tests pass


nhniina:
Indentations are sometimes off and we could still discuss with other SpineOpt developers what is the best way to group the outputs. Let's also ensure that an issue is created about the problem with several reports linking to one output, so that the FIXME lines don't get lost.

PR #1117, opened 2024-10-11
fix fom_cost formulation 

nnhjy:
`fom_cost` needs to work in three cases:
1. the quantity of a `unit` is only given by `number_of_units` without investment availability
2. the quantity of a `unit` is only determined by investment
3. a mix of 1&2

This PR extends the `fom_cost` formulation from covering cases 2&3 to case 1. 

## Checklist before merging
- [x] Documentation is up-to-date
- [x] Unit tests have been added/updated accordingly
- [x] Code has been formatted according to SpineOpt's style
- [x] Unit tests pass

PR #1123, opened 2024-10-29
Change indices function in the fixed_om_costs function

### LINKED ISSUE #1122: [bug] function `fixed_om_costs` throws and error when using representative periods
**Describe the bug**
There is an error in the model when setting the objective function after the changes in #1117 when using representative periods feature.

```log
Setting energy_model objective...ERROR: KeyError: key (unit = ATFR_wind-off-existing, stochastic_scenario = realization, t = 2018-02-06T23:00~(1 hour)~>2018-02-07T00:00) not found
```

datejada:
Using `units_invested_available_indices` in the function `fixed_om_costs` to avoid errors when using the representative periods feature.

Fixes #1122 

## Checklist before merging
- [x] Unit tests pass

### DIFF
--- a/src/objective/fixed_om_costs.jl
+++ b/src/objective/fixed_om_costs.jl
@@ -46,7 +46,7 @@ function fixed_om_costs(m, t_range)
             # an extended period of time with a weight >=1, e.g. a representative month represents 3 months.
             * duration(t) 
             for (u, ng, d) in indices(unit_capacity; unit=indices(fom_cost))
-            for (u, _n, _d, s, t) in unit_flow_indices(m; unit=u, node=ng, direction=d, t=t_range);
+            for (u, s, t) in units_invested_available_indices(m; unit=u, t=t_range);
             init=0,
         )
     )
--- a/src/objective/fixed_om_costs.jl
+++ b/src/objective/fixed_om_costs.jl
@@ -46,7 +46,14 @@ function fixed_om_costs(m, t_range)
             # an extended period of time with a weight >=1, e.g. a representative month represents 3 months.
             * duration(t) 
             for (u, ng, d) in indices(unit_capacity; unit=indices(fom_cost))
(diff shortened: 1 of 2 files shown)


datejada:
@nnhjy I am putting back the indices of the function `fixed_om_costs` as `units_invested_available_indices` rather than `unit_flow_indices` as you changed in #1117. I am not sure why you changed the indices, but there must be a reason. Would you mind reviewing that what you want to achieve in your changes still works with these indices? If your changes work with this change, we should keep the indices as they were before so they work correctly with the representative periods feature.
Thanks!

nnhjy:
Thanks, @datejada! This reminds me of why `units_invested_available_indices` was used. (Probably it was also me who made it so the `fom_cost` could work with some cases with the representative period feature.) Now, the whole picture is clearer.

As mentioned in #1117, using `units_invested_available_indices` would overlook the non-invested units, i.e. those units defined simply by `number_of unit`. That was the very reason I changed the setting.

There needs a solution to cover both settings for a `unit` while allowing for the the representative. I can give a try to fix that if you don't mind working with this spontaneous branch this week.

datejada:
> Thanks, @datejada! This reminds me of why `units_invested_available_indices` was used. (Probably it was also me who made it so the `fom_cost` could work with some cases with the representative period feature.) Now, the whole picture is clearer.
> 
> As mentioned in #1117, using `units_invested_available_indices` would overlook the non-invested units, i.e. those units defined simply by `number_of unit`. That was the very reason I changed the setting.
> 
> There needs a solution to cover both settings for a `unit` while allowing for the the representative. I can give a try to fix that if you don't mind working with this spontaneous branch this week.

@nnhjy great! Thanks for the quick reply. You have the full picture 😄 so, it would be much appreciated if you could look at it. Please let me know if I can help you with this.

datejada:
> @datejada Would the latest commits resolve issue #1122 ?

Hi @nnhjy, yes, it does; thank you so much.

PR #1147, opened 2024-12-02
Fix temporal resolution connection_ratio_out_in

### LINKED ISSUE #1146: Fix flexible time resolution in constraint `fix_ratio_out_in_connection_flow`

gnawin:
Closes #1146

## Checklist before merging
- [x] Unit tests pass


gnawin:
I made some updates, could you have a further review @Tasqu? Thanks!

Another issue that I have not touched upon is the delays. Since delay basically introduces a separate resolution, we (@datejada and I) found out:
- if the delay is multiples of the resolutions from the node, it works. 
- otherwise, it does not. What would be desired is that the constraint indices are the highest resolution between the nodes and the delay. If we decide to go that way, any idea of how the implementation works?

Tasqu:
> Another issue that I have not touched upon is the delays. Since delay basically introduces a separate resolution, we found out:
> 
> * if the delay is multiples of the resolutions from the node, it works.
> * otherwise, it does not. What would be desired is that the constraint indices are the highest resolution between the nodes and the delay. If we decide to go that way, any idea of how the implementation works?

This might be tricky. I took a quick look, and I'm not sure why it wouldn't be working even if the delay isn't an exact match. My best guess is that it might have something to do with the `to_time_slice` and `_to_time_slice` functions, as those are used to figure out which time slice the delay should be pointing at.

Unfortunately, I most likely don't have time to look into this any further for now either :(

gnawin:
We found out that the currently implementation does not scale well with a large model. It was discovered because this is causing a problem in the test for economic representation. This issue is not related to the economic representation itself, but because this case has the largest model in the dataset (hourly for a year). Here are the observations.

1. By setting the model to hourly for a day, the resource usage is:
   [constraint_fix_ratio_out_in_connection_flow]  4.107311 seconds (6.92 M allocations: 359.169 MiB, 1.23% gc time, 98.64% compilation time)

    which is among the largest resource eaters, by comparable to e.g., 
    [constraint_fix_ratio_out_in_unit_flow]  4.330690 seconds (7.32 M allocations: 363.061 MiB, 1.10% gc time, 99.43% compilation 
    time)

   far more than most , e.g.,
   [constraint_fix_ratio_out_out_unit_flow]  0.005151 seconds (730 allocations: 41.859 KiB, 97.33% compilation time)
   
2.  Notice that after the 1st run (i.e., after the complication), the usage drops to:
   [constraint_fix_ratio_out_in_connection_flow]  0.062868 seconds (659.58 k allocations: 50.421 MiB, 24.07% compilation time)

3. After the complication, by setting the model to hourly for a month, the resource usage is:
   [constraint_fix_ratio_out_in_connection_flow] 39.678608 seconds (597.19 M allocations: 44.108 GiB, 13.65% gc time, 0.05% compilation time)
   [constraint_fix_ratio_out_in_unit_flow]  0.334825 seconds (3.56 M allocations: 286.618 MiB, 5.16% gc time, 3.65% compilation time)
   [constraint_fix_ratio_out_out_unit_flow]  0.000068 seconds (185 allocations: 15.609 KiB)

In summary, the implementation in this PR has a scalability problem, using both lots of time and memory. Maybe it is caused by type instability or other reasons, will investigate further.

src/constraints/constraint_ratio_out_in_connection_flow.jl
```diff
@@ -111,17 +110,16 @@ end
 
 function constraint_ratio_out_in_connection_flow_indices(m::Model, ratio_out_in)
```
gnawin, line 111: Hi @Tasqu @DillonJ @manuelma  

I want to get the highest resolution between t_out and t_in (with delays), with their associated paths.
But this code does not return the constraints I want, do you see something weird in this indices function?

src/constraints/constraint_common.jl
```diff
@@ -51,6 +51,23 @@ function t_lowest_resolution_path(m, indices, extra_indices...)
     ((t, path) for (t, scens) in scens_by_t for path in active_stochastic_paths(m, scens))
 end
 
+function t_highest_resolution_path(m, indices, extra_indices...)
+    isempty(indices) && return ()
+    if length(stochastic_scenario()) == 1
+        s = only(stochastic_scenario())
+        return ((t, [s]) for t in t_highest_resolution!(m, unique(x.t for x in indices)))
+    end
+    scens_by_t = t_highest_resolution_sets!(m, _scens_by_t(indices))
```
gnawin, line 60: This line returns only the highest t and its scenario, but not the overlapping scenario. e.g., flow a with parent and hourly resolution (say hour 1 with parent, and hour 2 with parent), flow b with child with two-hourly resolution (say hour 1-2 with child), I'd like to get the timeslice with hourly resolution for both parent and child (hour 1 with parent and child, hour 2 with parent and child), however, the line only returns parent of that timeslice (hour 1 with parent and hour 2 with parent). This means I'm losing the indice of flow b, and therefore this variable is missing from the constraint (e.g., hour 1).

src/constraints/constraint_ratio_out_in_connection_flow.jl
```diff
@@ -111,17 +110,19 @@ end
...
         if !_has_simple_fix_ratio_out_in_connection_flow(conn, ng_out, ng_in)
-        for (t, path_out) in t_lowest_resolution_path(
+        for (t_out, path_out) in t_highest_resolution_path(
             m, connection_flow_indices(m; connection=conn, node=ng_out, direction=direction(:to_node))
-        )
-        for path in active_stochastic_paths(
-            m, 
-            Iterators.flatten(
+        ) 
+        for (t, path) in t_highest_resolution_path(
```
gnawin, line 119: I'd like to get the highest resolution with the associated path. For this:
- I use 2 loops here, the first one (line 116) to get the highest among all the outflows, and the second one (line 119) to get the highest among the outflows and the inflows. Note the reason to use 2 loops instead of 1 is that I need t_out and s_out to get the indices for the inflows.
- Please also check if the way I incorporate the outflows and the corresponding inflows make sense, i.e., I get the t_out and path_out from the first loop, and I use t_out directly in the second loop and get all the scenarios because they can be multiple scenarios but only 1 timeslice.

datejada: Review again if we can use `active_stochastic_paths` in the second `for` (or third?) loop such that we guarantee that we get the right stochastic path.

Tasqu (approved):
Looks good to me, nice job 👍

Tasqu:
@gnawin it seems like there are still some conflicts with `master` regarding the examples preventing a merge. I'm guessing it's simply related to the added `use_highest_resolution_constraint_ratio_out_in_connection_flow` parameter definition.

As far as the main changes go, they look ok to me, so if you're feeling confident go ahead and merge this pull request after addressing the example conflicts.

src/constraints/constraint_common.jl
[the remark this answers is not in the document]
gnawin: I guess this comment is outdated.

PR #1149, opened 2024-12-03
Add options to improve model creation speed

### LINKED ISSUE #1148: Add option `set_string_names_on_creation` in `run_spineopt`
The option `set_string_names_on_creation` set to `false` reduces the time to create the model at the cost of not having the names in the model available after solving. Most of the cases the user don't need the names, however, it is useful to have the names during a debugging stage. That is why it should be an option for the user.

More information here: https://jump.dev/JuMP.jl/stable/tutorials/getting_started/performance_tips/#Disable-string-names

datejada:
Following the performance tips from [JuMP docs](https://jump.dev/JuMP.jl/stable/tutorials/getting_started/performance_tips/#Performance-tips), this PR implements the following option in the `run_spineopt` function:

- `use_model_names`(default = `true`): whether or not to create the model with names
- `add_bridges`(default = `false`): whether or not bridges from JuMP to the solver should be added to the model

**Notes**

- The `use_model_names` default is equal to `true`, so the users can gain more speed when creating the model by setting this option to `false`.

- The `add_bridges` option is set to `true` for the MGA algorithm, since it is necessary to run that feature.

Fixes #1148 

## Checklist before merging
- [x] Unit tests pass

### DIFF
--- a/docs/make.jl
+++ b/docs/make.jl
@@ -36,6 +36,7 @@ pages = [
         "Installation" => joinpath("getting_started", "installation.md"),
         "Verify installation" => joinpath("getting_started", "recommended_workflow.md"),
         "Troubleshooting" => joinpath("getting_started", "troubleshooting.md"),
+        "Performace tips" => joinpath("getting_started", "performance_tips.md"),
     ],
     "Tutorials" => Any[
         "Webinars" => joinpath("tutorial", "webinars.md"),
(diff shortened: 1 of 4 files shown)


datejada:
These are the results from the benchmark.

**Results disabling the name creation**
```
 BenchmarkTools.Trial: 3 samples with 1 evaluation.
 Range (min … max):  60.068 s … 65.056 s  ┊ GC (min … max): 18.35% … 21.78%
 Time  (median):     64.747 s             ┊ GC (median):    18.89%
 Time  (mean ± σ):   63.290 s ±  2.795 s  ┊ GC (mean ± σ):  19.71% ±  1.85%

 Memory estimate: 44.01 GiB, allocs estimate: 691724795.
``` 

**Results with the name creation**
```
 BenchmarkTools.Trial: 3 samples with 1 evaluation.
 Range (min … max):  67.082 s … 78.006 s  ┊ GC (min … max): 19.92% … 18.47%
 Time  (median):     71.319 s             ┊ GC (median):    18.73%
 Time  (mean ± σ):   72.135 s ±  5.508 s  ┊ GC (mean ± σ):  18.99% ±  0.78%

 Memory estimate: 44.96 GiB, allocs estimate: 714989451.
```

datejada:
Changes commented during developers meeting

PR #1154, opened 2025-01-17
Use `align_variable_duration_unit` function instead of `convert_to_days`

### LINKED ISSUE #1152: Improve convert_to_days function
https://github.com/spine-tools/SpineOpt.jl/blame/f6bf05a12d80f96e7a2f3d301adf07b448bd9b41/src/constraints/constraint_common.jl#L74

- [ ] Define the function outside of the past indices function
- [ ] Add to the docs the assumption of this function

Double-check if we can use this function instead:

https://github.com/spine-tools/SpineOpt.jl/blob/f6bf05a12d80f96e7a2f3d301adf07b448bd9b41/src/util/misc.jl#L179

datejada:
Reusing the function `align_variable_duration_unit` instead of creating a new one `convert_to_days`

Fixes #1152

## Checklist before merging
- [x] Unit tests pass

### DIFF
--- a/src/constraints/constraint_common.jl
+++ b/src/constraints/constraint_common.jl
@@ -75,7 +75,7 @@ function _past_indices(m, indices, param, s_path, t; kwargs...)
         (;
             ind...,
             weight=ifelse(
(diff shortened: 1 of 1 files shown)


datejada:
@nnhjy your function does the same as the one I added before. So, I am reusing yours since it is better documented and to avoid having dirty hacks inside the functions. Thanks!

PR #1170, opened 2025-02-28
Merge changes from v0.7.2.x into Master

### LINKED ISSUE #1108: Representative periods not working with variables requiring history
**Describe the bug**
If the unit_state_transion constraint is activated because a min up or down time is defined, the model fails due to key error where the variable is not defined for the history timeslice

### LINKED ISSUE #1124: Couple TulipaClustering for representative periods
Related:
- #1108

### LINKED ISSUE #1136: Representative periods with variable replacement causing errors
**Describe the bug**
It seems that when e.g. connection_flow or others which have **replacement expressions**, are present with **representative periods**, errors can emerge. This is because the variable indices provided to `_representative_periods_mapping()` include also the variables which are replaced.

**To Reproduce**
Add a connection with `fix_ratio_out_in_connection_flow`. Add representative periods as explained in [How to set up representative days for investment problems](https://spine-tools.github.io/SpineOpt.jl/latest/how_to/set_up_representative_days_for_investment_problems).

- Which SpineOpt version? _latest master_

**Possible solution**
Adding the following to the end of `add_variable()` may solve it. I'm not sure it it interferes with replacement expressions.

```
function add_variable!(
.
.
.

    function corrected_indices(m; kwargs...)
        return [ind for ind in indices(m; kwargs...) if !haskey(replacement_expressions, ind)]
    end

    isempty(SpineInterface.indices(representative_periods_mapping)) || merge!(
        vars, _representative_periods_mapping(m, vars, corrected_indices)
    )
```

### LINKED ISSUE #1139: Representative periods with constraint_ratio_out_in_connection_flow causing problems
**Describe the bug**
The constraint `constraint_ratio_out_in_connection_flow` causes problems when representative periods are used. That is, when the constraint is present in the first place. Often it is removed by `_has_simple_fix_ratio_out_in_connection_flow()` – in this case I artificially make sure that it will be added by returning false from this function.

What seems to happen when representative periods are used, the function `_past_connection_input_flow_indices()` produces indices both from the **original** temporal block associated with a node and the **representative period** temporal block when the same time slice exists in both. This totally messes up the constraint, producing weird results.

**To Reproduce**
- make sure the constraint is added by returning false from _has_simple_fix_ratio_out_in_connection_flow()
- use the example model in [aug_elexia.json](https://github.com/user-attachments/files/17771618/aug_elexia.json)

- Which SpineOpt version? latest master

**suggested fix**
Remove` temporal_block=anything` in call to `connection_flow_indices()` in  `_past_connection_input_flow_indices()`

tarskul:
Merging significant changes from v0.7.2.x to the master branch:

- Improvements on Benders decomposition
- Fixes and improvements in representative periods (new! possibility to use fractional weights)
- Add Monte Carlo feature
- Improvements on the Multi-stage feature
- Performance Improvements

Fixes #1124 
Fixes #1136
Fixes #1139 
Fixes #1108

## Checklist before merging
- [ ] Unit tests pass


src/variables/variable_units_invested_available.jl
```diff
@@ -57,7 +61,6 @@ function add_variable_units_invested_available!(m::Model)
         lb=constant(0),
         int=units_invested_available_int,
         fix_value=fix_units_invested_available,
-        internal_fix_value=internal_fix_units_invested_available,
         initial_value=initial_units_invested_available,
```
manuelma, line 64: This needs to be `initial_value=_initial_units_invested_available` - note the sneaky leading underscore - to use the function defined above in line 47. Otherwise the unit investment variable is not fixed to zero at the beginning of the horizon. The same should be donde for storages and connections in the respective files. This should fix Benders tests.

To explain, the function `_initial_units_invested_available` defined on line 47 ensures that the initial value is either the value defined by the user via the `initial_units_invested_available` parameter, or zero if no value defined.

Nice work @DillonJ @tarskul this seems to be coming along nicely. Sorry for the merge issues, most of them must be my fault...

DillonJ: Thanks  @manuelma it wasn't actually that difficult to resolve the conflicts and I think we all equally contributed to the issues :-) I've made that fix - let's see how the tests go!

tarskul:
I don't immediately see errors in the tests now, but for some reason they seem to get stuck. @datejada can you run the tests locally on your computer again to exclude github shennanigans?

PR #1172, opened 2025-02-28
Add functionality for timedependent forced outage

mihlema:
In order to account for time-dependent outage rates (e.g., influenced by weather, see for example: [Murphy - A time-dependent model of generator failures and recoveries captures correlated events and quantifies temperature dependence](https://www.sciencedirect.com/science/article/pii/S0306261919311870), it'd be nice to be able to support MTTF and MTTR as timeseries parameters. The adaptation for this script supports this, with the caveat that timeseries currently only supports Float values. The script will treat Floats as number of Days for respective parameters. 

Other than minor update to write_parameter function (was outdated) and added the keyword "on_conflict"=replace. The latter one seemed reasonable as I was testing and noticed that I created "more and more" outages, as the script would append the timeseries, leading to erroneous timeseries for forced outages. I left it as an optional keyword, In case e.g. do want to append an existing timeseries (e.g., originally created for year x, and now appended for year y, as long as they don't overlap)

Fixes # (no issue addressed)

## Checklist before merging
- [ ] Unit tests have been added/updated accordingly - no unit test available for this script yet. This should however be established in the future
- [ ] Code has been formatted according to SpineOpt's style - probably not yet, there's a few comments reviewers need to check out first before
- [ ] Unit tests pass (N/A)

### DIFF
--- a/src/generate_forced_outages.jl
+++ b/src/generate_forced_outages.jl
@@ -19,36 +19,41 @@
 using Distributions
 using Random
 
-function _rand_time(mean_time; resolution)
+function _rand_time(mean_time, u; resolution)
+    mean_time = (typeof(mean_time) == Float64 ? Day(mean_time) : mean_time)
     mean_time = round(mean_time, resolution(1))
     resolution(round(Int, rand(Exponential(iszero(mean_time.value) ? 1e-6 : mean_time.value))))
 end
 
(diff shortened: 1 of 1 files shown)


DillonJ:
@mihlema This would be great and it's nice that you're complimenting the existing forced outage functionality. It would be great if we could fix the conflicts and merge

PR #1175, opened 2025-03-03
Revert "Merge changes from v0.7.2.x into Master"

nnhjy:
Reverts spine-tools/SpineOpt.jl#1170


datejada:
@nnhjy and @tarskul since it is merged into master, then I would prefer to create a separate issue for to fix the test for the representative periods

PR #1179, opened 2025-03-05
Attempt to fix/resolve the representative_periods tests issues

### LINKED ISSUE #1176: Bug: Representative periods tests are not working
**Describe the bug**
The tests get stuck at this line and never finish...

https://github.com/spine-tools/SpineOpt.jl/blob/200d2d1ab5099e226d5b2d79102386899a2f23d1/test/run_spineopt_representative_periods.jl#L199

I tested using Gurobi instead of HiGHS and it is the same, meaning that it is in the SpineOpt side and not the solver side.

The tests have been disabled temporarily in the master branch, see commit https://github.com/spine-tools/SpineOpt.jl/commit/942a36a3ae2357191c6fbfda6a2354c7ee24cd50

this part of the work in PR #1170 

@manuelma sorry to bother you again, do you have any idea what might the problem might be? ANy hint will help. Thanks!

cc: @tarskul @nnhjy @DillonJ 

**To Reproduce**

Run the tests in the file https://github.com/spine-tools/SpineOpt.jl/blob/master/test/run_spineopt_representative_periods.jl

**Expected behavior**
The test should pass 😃 

**Screenshots**
N.A.

- Which SpineOpt version? latest master
- Which toolbox version? N.A.
- OS: it happens in all OS

**Additional context**
N.A.

nnhjy:
1. draft unit_tests for `min_up/down_time` constraints
2. prevent activating multi-threading when it is not intended

Fixes #1176

## Checklist before merging
- [x] Unit tests pass

### DIFF
--- a/src/run_spineopt_basic.jl
+++ b/src/run_spineopt_basic.jl
@@ -920,17 +920,17 @@ function _calculate_duals_fallback(m; log_level=3, for_benders=false)
     reduced_cost_fallback(var) = ReducedCostPromise(ref_map[var])
     _save_marginal_values!(m, dual_fallback)
     _save_bound_marginal_values!(m, reduced_cost_fallback)
-    if isdefined(Threads, Symbol("@spawn"))
-        task = Threads.@spawn @timelog log_level 1 "Optimizing LP..." optimize!(m_dual_lp)
-        lock(m.ext[:spineopt].dual_solves_lock)
-        try
-            push!(m.ext[:spineopt].dual_solves, task)
-        finally
-            unlock(m.ext[:spineopt].dual_solves_lock)
-        end
-    else
-        @timelog log_level 1 "Optimizing LP..." optimize!(m_dual_lp)
-    end
+    # if isdefined(Threads, Symbol("@spawn"))
+    #     task = Threads.@spawn @timelog log_level 1 "Optimizing LP..." optimize!(m_dual_lp)
+    #     lock(m.ext[:spineopt].dual_solves_lock)
+    #     try
+    #         push!(m.ext[:spineopt].dual_solves, task)
+    #     finally
+    #         unlock(m.ext[:spineopt].dual_solves_lock)
+    #     end
+    # else
+    #     @timelog log_level 1 "Optimizing LP..." optimize!(m_dual_lp)
+    # end
 end
 
 function _relax_discrete_vars!(m::Model, ref_map::ReferenceMap; fix_variables=())
--- a/src/run_spineopt_basic.jl
+++ b/src/run_spineopt_basic.jl
@@ -925,9 +925,11 @@ function _calculate_duals_fallback(m; log_level=3, for_benders=false)
     # that determines whether multi-threading is activated (has a value) or not (no value) in Julia.
     # isdefined(Threads, Symbol("@spawn")) is always true in the current version of Julia (1.11) 
     # no matter whether multi-threading is activated or not i.e. Threads.nthreads()=1 by default.
-        println("Having a `println` here allows the LP optimization to be done despite suspending on its finishing.")
+        #TODO: This command would suspend the running of `m = run_spineopt(...; optimize=true, ...)`
+        # in the unit test `run_spineopt_representative_periods.jl`. Suspension comes at launching the `optimize!()`.
+        # Add an arbitraty command, either before or after this command, could shift the suspension 
+        # to the completion of the `optimize!()`.
         task = Threads.@spawn @timelog log_level 1 "Optimizing LP..." optimize!(m_dual_lp)
-        println("The original `task=...` command suspends before the LP optimization begins.")
         lock(m.ext[:spineopt].dual_solves_lock) 
         try
             push!(m.ext[:spineopt].dual_solves, task)
--- a/src/run_spineopt_basic.jl
+++ b/src/run_spineopt_basic.jl
@@ -930,7 +930,7 @@ function _calculate_duals_fallback(m; log_level=3, for_benders=false)
         # Add an arbitraty command, either before or after this command, could shift the suspension 
         # to the completion of the `optimize!()`.
         task = Threads.@spawn @timelog log_level 1 "Optimizing LP..." optimize!(m_dual_lp)
-        lock(m.ext[:spineopt].dual_solves_lock) 
+        lock(m.ext[:spineopt].dual_solves_lock)
         try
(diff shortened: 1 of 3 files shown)


DillonJ:
Hi @nnhjy 

I've just commited a change to master to do the same thing - except, I have a suspicion that the reason the multithreading didn't work for this test was because of the missing min updown time tests and how the tests in run_spineopt_represenstative_periods.jl have been implemented. They rely on function multiple dispatch based on the value of the contraint value... I think the missing function definitions for min up/down times may have caused the issue. We will see shortly if tests pass for my change. If this works, it's not necessary to do anything with the multithreading - which is very useful.

test/run_spineopt_representative_periods.jl
```diff
@@ -468,6 +468,49 @@ function _expected_representative_periods_constraint(
...
+    # min_up_time of unit "h2_gen" is implicitly set to be the default model duration unit in preprocess_data_structure.jl, 
+    # triggered by setting "online_variable_type" to be "unit_online_variable_type_integer" in the test dataset.
+    s = only(s_path)
+    @fetch units_on, units_started_up = m.ext[:spineopt].variables
+    @build_constraint(
+        units_on[u, s, t]
+        >=
+        sum(
+            units_started_up[u, s_past, t_past] * weight
+            for (u, s_past, t_past, weight) in SpineOpt.past_units_on_indices(m, min_up_time, u, s_path, t)
```
DillonJ, line 488: Is this cheating a little? If there's an error with SpineOpt.past_units_on_indices this test won't detect a problem. We should test against what we know the constraint should be, no?

test/run_spineopt_representative_periods.jl
```diff
@@ -468,6 +468,49 @@ function _expected_representative_periods_constraint(
...
+    # triggered by setting "online_variable_type" to be "unit_online_variable_type_integer" in the test dataset.
+    s = only(s_path)
+    nou = vals["unit", string(u), "number_of_units"]
+    @fetch units_invested_available, units_on, units_shut_down = m.ext[:spineopt].variables
+    @build_constraint(
+        nou + units_invested_available[u, s, t_invest] - units_on[u, s, t] 
+        >= 
+        sum(
+            units_shut_down[u, s_past, t_past] * weight
+            for (u, s_past, t_past, weight) in SpineOpt.past_units_on_indices(m, min_down_time, u, s_path, t)
```
DillonJ, line 510: Same as above, are we cheating using SpineOpt.past_units_on_indices?

Tasqu:
@nnhjy You made a powerful observation about this testset: it's not necessary to optimize the problem at all, which is what currently causes the freeze. Simply giving `optimize=false` bypasses the problematic part as long as we fix the missing constraint tests.

We really shouldn't optimize the problem in the tests unless necessary to save time anyway. It's the original reason the `optimize=false` keyword was initially added 😆

Tasqu:
Like I mentioned in my comment in #1176, based on my further investigation this freeze seems to me like an arbitrary multi-threading timing issue. Hopefully this is caused by the test model completing faster than expected by the multi-threading, resulting in complications.

If we're lucky, this is a really rare timing bug that never occurs with any actual models. _(All our other unit tests also pass through this without freezing)_

Tasqu:
@nnhjy seems like the unit test fixes by @DillonJ in latest master actually pass, as long as optimization is bypassed. The `run_spineopt` in this testset still hangs for an eternity, though.

My suggestion is that we re-enable the representative period tests and simply pass `run_spineopt(...; optimize=false)` for now. Based on my testing, this freeze appears almost arbitrary, and hopefully will only ever occur in trivial test systems and never in any actual models.

nnhjy:
Many thanks @DillonJ for https://github.com/spine-tools/SpineOpt.jl/pull/1179#issuecomment-2700557757

Indeed, both our commits do the similar thing in a little different manner. I think the `min_up/down_constraint`s are automatically generated, triggered by the implicit creation of parameter `min_up/down_time`="model default unit duration" (corresponding tests are added in this PR) for the unit `h2_gen` because the testset defines its `online_variable_type` to be `integer/binary`. It would help for later diagnosis when needed to display this mechanism in the unit tests, I suppose. 

And having added the missing expected tests for `min_up/down_time` constraints (as done in both commits) fixes the failing tests associated with the two constraints @Tasqu . But the suspension issue seems to remain (as mentioned in https://github.com/spine-tools/SpineOpt.jl/issues/1176#issuecomment-2700554864). https://github.com/spine-tools/SpineOpt.jl/pull/1179#issuecomment-2700614146 sounds like an interesting point. We might give a try.

Actually, this PR does not cancel multithreading. Commit a0d9022419dd48805b2c4e6956065fc07628c6a4 is just to ensure using multitreading when it is really activated. For now, `julia` always has the `Threads.@spawn` defined no matter mutithreading is activated in the system or not. There are comments that clarify this in the code block.

test/run_spineopt_representative_periods.jl
nnhjy: I more or less feel the same as well. That's actually one reason for it not being committed to the `master` directly.

The main idea is to disclose more structural info. relevant to the constraint implementation here. To give the consequent appearance that we know at the moment definitely works, but that provides little clue on what is happening when one gets back a while later or to new developers.

I'm trying to figure out a more proper way to handle this. Any thoughts?

nnhjy:
@Tasqu Based testing on my local machine, setting `optimize=false` bypasses the suspension issue.
The original `run_spineopt()` function sets `optimize=true` by default. That's why we saw the freezing run in the tests for  `representative_periods` as it originally used `run_spineopt()` without specifying the `optimize` keyword.

There seems another way to avoid the suspension in this latest commit dd11e5ca2ecc8f04c481fc55cfd97616ac901420, which could in the meantime avoid the concerns in https://github.com/spine-tools/SpineOpt.jl/issues/1176#issuecomment-2700605650.

src/run_spineopt_basic.jl
Tasqu, line 1: I like the comment explaining the situation 👍

Tasqu (approved):
I especially like the comment explaining the annoying freeze. That's good to have for the future if we happen to run into this again.

As for the unit tests, I didn't really have the time to look into them in detail. They seem fine on a glance, and we can always tweak them later if we need to.

Tasqu:
Hopefully didn't merge this too prematurely, but @gnawin needs a "working" master for his temporal resolution fixes. We can always fix this in post if issues or concerns emerge.

PR #1188, opened 2025-03-17
Set default values for economic parameters

gnawin:
Give default values for economic parameters to ensure that they can be saved without errors.

## Checklist before merging
- [ ] Unit tests pass

### DIFF
--- a/src/data_structure/economic_structure.jl
+++ b/src/data_structure/economic_structure.jl
@@ -130,6 +130,12 @@ function generate_capacity_transfer_factor!(m::Model, obj_cls::ObjectClass, econ
     invest_temporal_block = economic_parameters[obj_cls][:set_invest_temporal_block]
     param_name = economic_parameters[obj_cls][:set_capacity_transfer_factor]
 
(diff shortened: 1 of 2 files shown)


datejada (commented):
Thanks for the changes and the fix. Can I ask you please to add a test to cover this in the future? I mean, add a test that has the economic representation option but does not have the outputs defined. So, we can catch the error from the tests. Thanks!

datejada:
> @datejada actually the problem is not about defining the outputs in the reports or not, but saving results will always error if the economic parameters do not have default values.
> 
> I agree with you that we should have a test to somehow cover that, but I did not find a direct way to test that since setting default values are deep in the code and that removing the default values in the test did not seem doable.
> 
> Instead, I test indirectly. Knowing not having defaults will have a problem, I test whether we have enough parameter definitions. This is not a perfect way to do it, but at least it shows in the test that we should pay attention to this in the future. Let me what you think!

I see, it makes sense, thanks for the effort, and let's leave it like this 😃

datejada (approved):

PR #1192, opened 2025-04-08
Update migration scripts and the template version number

nhniina:
Change spineopt_template.json version number to 16 and update the migration scripts so that the newest template is loaded.

## Checklist before merging
- [x] Unit tests pass

### DIFF
--- a/examples/capacity_planning.json
+++ b/examples/capacity_planning.json
@@ -3403,7 +3403,7 @@
         [
             "settings",
             "version",
-            15,
+            16,
             null,
             "Current version of the SpineOpt data structure. Modify it at your own risk (but please don't)."
         ],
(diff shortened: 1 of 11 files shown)


tarskul:
Ok, I think we can ignore the codecov. I looked into the report and it just systematically skips the lines with the value 'true'.

For example:
```
function _run_migration(url, version, log_level)
	upgrade_fn = get(_upgrade_functions, version, nothing)
	upgrade_fn === nothing && return false
	upgrade_fn(url, log_level) || return false
	true
end
```

In that function it says that the lines with 'upgrade' are covered but the line with 'true' isn't, which does not make sense at all.

And since the function you added only has the value true in it, it does not consider the function.

tarskul:
Yeah, I'm rerunning the test in the hopes that it will eventually pass. But even if it keeps failing, it cannot possible have anything to do with your changes so we'll merge it anyway. Though we may have to add an issue about the failing test.

PR #1209, opened 2025-05-19
Add min unit flow constraint

### LINKED ISSUE #1203: Add a constraint for the minimum unit flow limit
Similar to #1201 but for nodes. Some things to consider:

1. Can we use the `minimum_operating_point`?
2. How those two constraints interact can be both at the same time, or should we avoid using them simultaneously (they can conflict and cause infeasibilities)

datejada:
- Add parameter `unit_min_factor`
- Add calculated parameter (internal) `unit_flow_lower_limit`
- Add constraint `unit_min_flow`

Fixes #1203

## Checklist before merging
- [ ] Unit tests pass


datejada:
Not adding this constraint, see comments on issue #1203

PR #1245, opened 2025-08-06
Introduce has_free_start and use it to create blocks private history

### LINKED ISSUE #1238: Allow discontinuity in operational variables between timeslices
This relates to #1237 

@aporrasc is trying to set up an investment planning model. He is trying to create a model that consists of multiple snapshot years in a single optimisation. He is trying to do this with temporal blocks representing each snapshot year. One for investments, and one for operations, for each snapshot year. E.g:
TB_Ops_2030: block_start = 01/01/2030, block_end = 01/01/2031, resolution: 12h
TB_Ops_2040: block_start = 01/01/2040, block_end = 01/01/2041, resolution: 12h
TB_Investments_2030: block_start = 01/01/2030, block_end = 01/01/2031, resolution: 1Y
TB_Investments_2040: block_start = 01/01/2040, block_end = 01/01/2041, resolution: 1Y

[@aporrasc I'm leaving out the 10Y thing as I think it's a bit of a complication, doing it as above reduces the problem to the single issue of operational variable continuity]
 
The problem is that we need to free up the node state variables between the two operational temporal blocks. In this case, we don't want continuity between the last timeslice of 2030 and the first of 2040. We wants a free history timeslice just before 2040 to free up the node state variables (and other variables) so the 2040 operational variables have a "fresh start". Because we are solving this as a single optimisation, we only have free variables preceding 2030 but not before 2040. We don't want the model to roll because we want the investment decisions to see all the snapshot years at once.
 
For example, the cyclic condition constraint for the 2030 TB will constrain the node_state in the last timeslice of 2030 relative to the last history timeslice (timslice immediately preceding 2030). But for the 2040 TB, cyclic condition will constrain the node_state of last timeslice of 2040 relative to the last timeslice of 2030 creating a constraint that will overly constraint investments and operations. We effectively want a free variable in between the end of 2030 and the beginning of 2040 to free up the operational variables so there is no continuity between the 2030 operational variables and the 2040 operational variables (but we do want continuity for the investment variables).
 
An idea I had was to have a couple of TB parameter like "operational_variables_free" and "investment_variables_free" or something like that. Then you can insert such a TB in between 2030 and 2040 and it would create a sort of middle history window. You would just need to deactivate constraints and bounds, just like the history variables... but only for operational variables.
 
This is just an idea, maybe someone can think of a better way? If there was an easy way to create some timeslices in the intervening period and include them in the history timeslices collection, would that do it? Does anyone have other ideas how this could be done relatively easily?

@datejada @Tasqu @nnhjy @nhniina @gnawin @manuelma

manuelma:
Here we introduce the has_free_start boolean parameter for the temporal_block class, allowing us to setup a free start for operational variables in the middle of the optimisation.

The implementation relies on creating history time slices directly before the blocks with has_free_start = true. This should (if I'm not mistaken) take care of everything because when creating constraints, we already skip time slices marked as history. So effectively, variables will be free for those history time slices just before the block, and the block will thus have a free start. 

I have added a test that checks that has_free_start is correctly handled when creating the temporal structure. What I believe is missing is a test that checks that such temporal structure actually satisfies the requirement for the variables to have a free start in the blocks with has_free_start = true. But @datejada offered to help with that (thanks). To clarify, what the test needs to do is build the model and check that no constraints are created for the history time slices that belong to the history of a block with has_free_start = true. And that the 'dynamic' constraints at the beginning of the block actually use the time slices from that history.

Fixes #1238 

Note: In the code I use the term 'subwindow' to refer to a block with free start, but we don't need to communicate that to the user of course. Also, before merging we should document the new parameter and perhaps somebody else could help with that?

## Checklist before merging
- [ ] Unit tests pass

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -41,3 +41,4 @@ Manifest.toml
 
 .vscode/
 debugging/
+deleteme.sqlite
--- a/examples/capacity_planning.json
+++ b/examples/capacity_planning.json
@@ -3608,6 +3608,13 @@
         ],
         [
             "temporal_block",
+            "has_free_start",
+            false,
+            "boolean_value_list",
+            "Whether the block requires a fresh start (own independent history)"
+        ],
+        [
+            "temporal_block",
             "is_active",
             true,
             "boolean_value_list",
--- a/examples/multi-year_investment_with_econ_discounting_consecutives.json
+++ b/examples/multi-year_investment_with_econ_discounting_consecutives.json
@@ -3431,6 +3431,13 @@
         ],
         [
             "temporal_block",
+            "has_free_start",
+            false,
+            "boolean_value_list",
+            "Whether the block requires a fresh start (own independent history)"
+        ],
+        [
+            "temporal_block",
             "is_active",
             true,
             "boolean_value_list",
--- a/examples/multi-year_investment_with_econ_discounting_milestones.json
+++ b/examples/multi-year_investment_with_econ_discounting_milestones.json
@@ -3503,6 +3503,13 @@
         ],
         [
             "temporal_block",
+            "has_free_start",
+            false,
+            "boolean_value_list",
+            "Whether the block requires a fresh start (own independent history)"
+        ],
+        [
+            "temporal_block",
             "is_active",
             true,
             "boolean_value_list",
--- a/examples/multi-year_investment_without_econ_discounting.json
+++ b/examples/multi-year_investment_without_econ_discounting.json
@@ -3424,6 +3424,13 @@
         ],
         [
             "temporal_block",
+            "has_free_start",
+            false,
+            "boolean_value_list",
+            "Whether the block requires a fresh start (own independent history)"
+        ],
+        [
+            "temporal_block",
             "is_active",
             true,
             "boolean_value_list",
--- a/examples/reserves.json
+++ b/examples/reserves.json
@@ -2152,6 +2152,13 @@
         ],
         [
             "temporal_block",
+            "has_free_start",
+            false,
+            "boolean_value_list",
+            "Whether the block requires a fresh start (own independent history)"
+        ],
+        [
+            "temporal_block",
             "is_active",
             true,
             "boolean_value_list",
--- a/examples/rolling_horizon.json
+++ b/examples/rolling_horizon.json
@@ -2152,6 +2152,13 @@
         ],
         [
             "temporal_block",
+            "has_free_start",
+            false,
+            "boolean_value_list",
+            "Whether the block requires a fresh start (own independent history)"
+        ],
+        [
+            "temporal_block",
             "is_active",
             true,
             "boolean_value_list",
--- a/examples/simple_system.json
+++ b/examples/simple_system.json
@@ -2152,6 +2152,13 @@
         ],
         [
             "temporal_block",
+            "has_free_start",
+            false,
+            "boolean_value_list",
+            "Whether the block requires a fresh start (own independent history)"
+        ],
+        [
+            "temporal_block",
             "is_active",
             true,
             "boolean_value_list",
--- a/examples/stochastic.json
+++ b/examples/stochastic.json
@@ -3438,6 +3438,13 @@
         ],
         [
             "temporal_block",
+            "has_free_start",
+            false,
+            "boolean_value_list",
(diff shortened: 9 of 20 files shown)


datejada:
@manuelma and update from my side:

1. I started adding a test for the node state with the new functionality here: https://github.com/spine-tools/SpineOpt.jl/pull/1245/files#diff-da1b3c717b30c2051f358a46bb60ea29ddbbc71ccfd2a406aea0f8f03513f7e9R1101. It is not finished yet, mainly because I have some questions about what is happening with the variables and constraints when having discontinuous temporal blocks—for instance, the initial node state. If there are discontinuous temporal blocks, then that constraint no longer applies to any block, not even the first one that is not free. So, what to do here? I would expect as a user to have the initial node state applied if it is not free and not used if it is free. But it might be subject to interpretations. What do you think?
1. On the same test, I have noticed that some historical time slides disappear once a discontinuous temporal block occurs. For example, without discontinuous blocks, there is history for the two-hourly block, but once there is a discontinuous temporal block, the history of that block in another node disappears. Is that intended?
1. After solving the questions and reviewing this first test, the idea is to add more tests on the other constraints that require historical time slices (e.g., minimum up time, investment, etc...)    
1. In addition, I have cherry-picked @Mastomaki's changes in the objective function. Thanks for that 😄 Without them, the model breaks when defining the objective function of the example `multi-year_investment_with_econ_discounting_milestones.json` with the new functionality. I am still testing this example. I hope to finish it early next week and then update the example (and the docs) with the new functionality (instead of the fake temporal block).

manuelma:
@datejada if you don't mind I would like to give it a shot to fix the issue with the middle history variables not being created, that ok with you?

manuelma:
@datejada I have pushed something now to fix the above issue. Things go better now so perhaps you can pick it up provided all tests pass? Maybe you @aporrasc could also test your model and see if it works as expected? Note that we need to set has_free_start = true for all the blocks where the history needs to be immediately before it starts, even if it starts together with the model.

gnawin:
@datejada In general, you would have a problem when your constraint involves t_before and t_after, such as storage, UC, and flow delays in connection flow constraint. So first of all, if we want to do a through test, these are the ones to check first. I think storage is the most relevant for long-term investment analysis.

In addition to these, there is another one that I encountered that forced me adding the extra timeslice in the multi-year tutorial. In the node injection constraint, you also need the extra timeslice to create the correct constraint for the first timeslice of the discontinous block. Please see the attached two lp files
[two_models.zip](https://github.com/user-attachments/files/22041857/two_models.zip)
, one with that extra timeslice and the other without, and you can compare the difference easily. This example is taken from one of the tutorials with milestones [this one](https://github.com/spine-tools/SpineOpt.jl/blob/master/examples/multi-year_investment_with_econ_discounting_milestones.json).

I think the way for the tests is to check the creation/existence of the first constraints of the discontinous block.

DillonJ:
> I don't think so @DillonJ - the behaviour when has_free_start = false (the default) shouldn't be altered at all by any of these changes. The fact that old tests pass is a reassurance.

Thanks @manuelma. So just for avoidance of doubt... a model with just a regular single rolling operations TB will still have a free start in the first rolling window, even if has_free_start is not specified?

manuelma:
> > I don't think so @DillonJ - the behaviour when has_free_start = false (the default) shouldn't be altered at all by any of these changes. The fact that old tests pass is a reassurance.
> 
> Thanks @manuelma. So just for avoidance of doubt... a model with just a regular single rolling operations TB will still have a free start in the first rolling window, even if has_free_start is not specified?

Yes, good question @DillonJ - it will have a free start even if has_free_start is not specified.

DillonJ:
@manuelma looks like we have some failing tests

manuelma:
> @manuelma looks like we have some failing tests

Yes, only some runs. Tests have become less deterministic somehow. I am reruning those that failed, eventually they will pass.

datejada:
@manuelma I checked the changes and the new test, thanks, it works as expected 😄 I am working on the minimum up time test in a discontinuous block, I hope to finish before the next meeting. So, I can update you all. After that, I will create a test for the flows delays, and update the multi-year investment tutorial. I think with that we can merge. I will keep you posted If I find something else.

PR #1247, opened 2025-08-12
Don't generate represented variable-expressions if not needed

### LINKED ISSUE #1246: Avoid unnecessary represented-variable expressions when is_longterm_storage = false
If `is_longterm_storage = false` for a node, we only track `node_state` in the representative periods -no nodal balance constraints are written for the __represented periods__ as they are not needed. For that reason, we don't generate any `node_state` _variables_ for the represented periods (because no constraint will use them). But we __do__ generate other variables such as `unit/connection_flow`, `units_on`, `units_started_up`, and `units_shut_down` (to name the ones I'm aware of) that are also not needed. This is adding considerably to the build time in my experiments.

manuelma:
Fixes #1246

## Checklist before merging
- [ ] Unit tests pass

### DIFF
--- a/test/run_spineopt_representative_periods.jl
+++ b/test/run_spineopt_representative_periods.jl
@@ -362,7 +362,6 @@ function _expected_representative_periods_constraint(
             - elec_demand(t=t_after)
         )
     else#if n == node(:h2_node)
-        fr_h2e = vals["unit__node__node", ["h2_gen", "h2_node", "elec_node"], "fix_ratio_in_out_unit_flow"]
         @build_constraint(
             + node_injection[n, s, t_after]
             ==
@@ -371,7 +370,7 @@ function _expected_representative_periods_constraint(
             - (1 / 24) * node_state[n, s, t_after]
             + (1 / 24) * node_state[n, s, t_before]
             + unit_flow[unit(:electrolizer), node(:h2_node), d_to, s, t_after]
-            - fr_h2e * unit_flow[unit(:h2_gen), node(:elec_node), d_to, s, t_after]
+            - unit_flow[unit(:h2_gen), node(:h2_node), d_from, s, t_after]
         )
     end
 end
--- a/test/run_spineopt_representative_periods.jl
+++ b/test/run_spineopt_representative_periods.jl
@@ -362,6 +362,12 @@ function _expected_representative_periods_constraint(
             - elec_demand(t=t_after)
         )
     else#if n == node(:h2_node)
+        unit_flow_from_node = if t_after in all_rt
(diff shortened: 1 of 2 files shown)


test/run_spineopt_representative_periods.jl
Tasqu, line 1: Why is the `fr_h2e` (`fix_ratio_in_out_unit_flow` ) parameter no longer necessary? Is it included via expressions now, or something else?

manuelma:
Hi @datejada thanks for that! Can you perhaps check the speedup in creating the dependant variables? That's the part where the optimisation was done. It looks like for this model that was much less of a bottleneck though...

DillonJ:
@datejada thanks for the benchmark. The speedup I quoted in the email of 7 minutes instead of 45 minutes was total build time, so I believe your original benchmark is directly comparable to that. I believe we had only one long term storage node in our model. Perhaps @aporrasc's model has more which may explain the differece in speed up. Are we sure that is_longterm_storage is true only for those nodes that really need it? It's not needed for short duration storage, or anything that can be optimised well in a single 24 hour period.

datejada:
> @datejada thanks for the benchmark. The speedup I quoted in the email of 7 minutes instead of 45 minutes was total build time, so I believe your original benchmark is directly comparable to that. I believe we had only one long term storage node in our model. Perhaps @aporrasc's model has more which may explain the differece in speed up. Are we sure that is_longterm_storage is true only for those nodes that really need it? It's not needed for short duration storage, or anything that can be optimised well in a single 24 hour period.

There are 8 storage with `is_longterm_storage=true` in that case study, related to either reservoirs, CO2 emissions, biomass stock, or large batteries (>24h). IT doesn't feel like a lot of long-term storages. You can double-check in the JSON file attached here https://github.com/spine-tools/SpineOpt.jl/pull/1247#pullrequestreview-3119613716

And to be fair, it is not a massive optimization model to either:

```log
A JuMP Model
├ solver: HiGHS
├ objective_sense: MIN_SENSE
│ └ objective_function_type: JuMP.AffExpr
├ num_variables: 105072
├ num_constraints: 269973
│ ├ JuMP.AffExpr in MOI.EqualTo{Float64}: 76453
│ ├ JuMP.AffExpr in MOI.GreaterThan{Float64}: 12695
│ ├ JuMP.AffExpr in MOI.LessThan{Float64}: 45334
│ ├ JuMP.VariableRef in MOI.EqualTo{Float64}: 1323
│ ├ JuMP.VariableRef in MOI.GreaterThan{Float64}: 81027
│ └ JuMP.VariableRef in MOI.LessThan{Float64}: 53141
└ Names registered in the model: none
```

But, still, it takes around 4.5min to build in the optimized version 🤷 (of course, first run..., but these use cases are single-shot optimizations, not for rolling horizon...)

PR #1253, opened 2025-09-08
Delta formulation and free starts

### LINKED ISSUE #1233: Reformulate formulation with representative periods in SpineOpt
Reformulate SpineOpt's current formulation of storage/node using the proposal **"Implicit - Min-Max bounds formulation"** in this reference:

https://arxiv.org/pdf/[2409.19079](https://arxiv.org/pdf/2409.19079)

In summary:

1. There are two types of storage level (SoC) variables, the inter- and the intra-temporal. Intra for those inside the representative period, and inter to keep track of the chronology of the SoC using the intra-period information and the mapping of the original periods to the representative periods.
2. The proposal introduces new variables to track the maximum and minimum changes in the SoC intra. That is an improvement that will help to keep the SoC within bounds along the recovered periods.

manuelma:
This PR introduces the delta formulation for the representative periods model. The main changes are as follows:

- When generating time-slices for represented blocks, we ignore the `resolution` parameter and we instead use the resolution of the `representative_periods_mapping` parameter. This is because we only need time-slices at mapping resolution to write the deltas.
- For represented periods, the only constraints are the `min_node_state` and `node_state_capacity` constraints for nodes with `is_long_term_storage=true`. No other constraint is generated outside of representative periods.
- In the above two constraints, the `node_state` is obtained from a JuMP expression corresponding to the delta equation.
- We don't generate any variables for represented periods.

Also we introduce temporal blocks with free start using middle histories, superseding https://github.com/spine-tools/SpineOpt.jl/pull/1254 (now closed).

Fixes #1233

## Checklist before merging
- [ ] Unit tests pass

### DIFF
--- a/test/run_examples.jl
+++ b/test/run_examples.jl
@@ -10,7 +10,7 @@ objective_function_reference_values = Dict(
     "multi-year_investment_without_econ_discounting.json" => 131520100.0,
     "capacity_planning.json" => 328503000.0,
     "stochastic.json" => 126964.2857142857,
-    "representative_periods.json" => 279438.4922720359,
+    "representative_periods.json" => 314610.23274110217,
     "multi_stage_model_tutorial.json" => 1.91746092844232e9,
 )
 


DillonJ:
Great stuff. @manuelma reports the following build times for a version of our SpineH2IRL model:

Before any imporvements: 45 min
Improved implementation (PR https://github.com/spine-tools/SpineOpt.jl/pull/1247): 7 min
Delta formulation (This PR): 3 min

@datejada @aporrasc : Can you guys test on the WP5 test model?

Issue #963, opened 2024-04-05
T4.1 Core model structure, decomp, and model templates (VTT)


clizbe:
- #613 
- Updates to model structure to improve usability
- Decomposition approaches
- Multi-instance modelling considering Decomposition
- Model templates

Tasqu:
Most of the feasible changes are made now, so closing this issue. Unfortunately, not all envisioned changes could be implemented, with #1101 remaining as a mode for further discussion.

Let's see if and when we get the opportunity to implement further improvements.

PR #1276, opened 2025-12-29
Datastructure update

Tasqu:
First stage of some SpineOpt input data structure revisions, aiming to simplify the structure.

The largest change is reworking the `unit__node__node` and its mess of parameters into a `unit_flow__unit_flow` compound class, hoping to make it more intuitive and eliminate redundant parameters. Similar compound classes were also implemented for other `unit__from_node` or `unit__to_node` relationships of the past. Other than that, the changes involve renaming a bunch of parameters and consolidating parameter value lists.

Going over all the changes here would be too much. Take a look at the `data_structure/versions/major_upgrade_1.jl` file for a comprehensive list.

NOTE! This branch requires the `#139_attempt4` branch of SpineInterface to work due to relying on compound classes previously unsupported. As such, coordinated merges and version bumps are required to avoid version conflicts.

Addresses #963 and #1101 but touches on a bunch of stuff.

## Checklist before merging
- [x] Unit tests pass
- [x] REQUIRES A VERSION BUMP UPON MERGE!

## Later todos
Ideally, `connection`s would be reworked similar to how units have been here, but that change needs to be handled later. Furthermore, there are a bunch of parameters that could still be clarified not only for connections.

Co-authored-by: Niina Helistö <niina.helisto@vtt.fi>
Co-authored-by: nnhjy <43530784+nnhjy@users.noreply.github.com>


Tasqu:
Reading your feedback I realized a potential problem, though: With these changes it's impossible to read, modify, and re-write SpineOpt input data using SpineInterface, because of how "compound classes" are broken for SpineInterface.

Technically, the process should be reversible, but I haven't implement it. The real solution would be to align the SpineInterface and Spine-DB-API internals anyway, but that we're even less likely to have the resources for.

nnhjy:
Makes sense at this point as the migration is embedded within the `run_spineopt(...)`. In practice, it is not fatal so long as it can also work with incomplete DBs.

> 2. The migration scripts were never really designed to work on incomplete DBs as far as I'm aware, so it's nice if they still function to some extent. The extended options to the migration I've implemented were primarily for migrating the .json examples and templates in the repo, and don't affect "regular" migration at the moment.

nnhjy:
Then this point would be more in concern. The inconsistency between the output and input DBs would likely lead to many complaints... but, of course, this is still WIP, just to raise the point here.

> No (at least not at this stage). The main motivation was to try and simplify the input data. Output changes were omitted to minimize impact. Furthermore, writing functional "compound classes" (relationships of superclasses) cannot be exported correctly from SpineInterface to Spine-DB-API at the moment. This would require fundamentally renovating SpineInterface internals.

Tasqu (commented):
It would be nice to have some unit tests for `dt_fixed_duration`

Tasqu (commented):
Again, we could use new unit tests for `connection_fixed_annual_cost`. It's also eventually planned to be called just `fixed_annual_cost` to avoid unnecessary duplication with the practically identical unit parameter. `storage_fixed_annual_cost` is named differently, because it has a slightly different meaning (cost per storage capacity, not flow capacity).

The plan is also to eventually change `unit` and `connection` fixed costs to `flow`-related parameters, as they are often defined per-capacity in sources. See #1101 

Also, at least some unit tests would be nice to check that the term appears in the objective when the parameter is defined.

nnhjy:
Would parameter name duplication `unit/connection_fixed_annual_cost` make indices filtering, e.g.
```julia
indices(capacity_per_connection; connection=indices(connection_fixed_annual_cost))
```
faster than not duplicating:
```julia
indices(capacity_per_connection; connection=indices(fixed_annual_cost))
```
?

nnhjy:
> unit tests for `dt_fixed_duration`

unit tests for `dt_fixed_duration` added

Tasqu:
> Would parameter name duplication `unit/connection_fixed_annual_cost` make indices filtering, e.g.
> 
> ```julia
> indices(capacity_per_connection; connection=indices(connection_fixed_annual_cost))
> ```
> 
> faster than not duplicating:
> 
> ```julia
> indices(capacity_per_connection; connection=indices(fixed_annual_cost))
> ```
> 
> ?

It can have an impact, which is why SpineInterface supports `indices(<parameter_name>, <class_name>)` to avoid searching unnecessary classes for indices.

nnhjy:
> It can have an impact, which is why SpineInterface supports `indices(<parameter_name>, <class_name>)` to avoid searching unnecessary classes for indices.

1. changed `connection_fixed_annual_cost` into `fixed_annual_cost`
2. explicitly claim the <class_name> when using `indices(<parameter_name>, <class_name>)` in the objective term expression `fom_costs` (for consistency and future renaming of the current `fom_cost` of `unit`)
3. added the 4-case unit tests on the fixed annual costs for all `unit`, `connection`, `storage

nnhjy:
Commit 0f0a62dc8c00afba7d5f96c0f14290b61dfdf278 fixed multi-threading crashes of running the `objective.jl` tests on my local machine. 
Hope that could provide some inspiring hints on the overall threading issue.

PR #1299, opened 2026-04-20
Add checks for ramping parameters

### LINKED ISSUE #1298: Ensure consistency between UC and ramping parameters
Start_up_limit should stay in the interval [minimum_operating_point, minimum_operating_point + ramp_up_limit],
Shut_down_limit should stay in the interval [minimum_operating_point, minimum_operating_point + ramp_down_limit]

Check if the interval holds, when start_up_limit or ramp_up_limit is used. 
Same for shut_down_limit or ramp_down_limit.

gnawin:
Fixes #1298 

- add to the existing check_data_structure(), this is after pre-processing, but before creating the actual problem.
- line 225-238 checks lower bound of start_up_limit and shut_down_limit, line 236-269 checks the upper bounds
- in case **any** of the parameters are not filled in, the default values will be provided in the check - this behavior is what happens during the constraint building. For example, if a user only gives ramp_up_limit or ramp_down_limit but **not** start_up_limit or shut_down_limit, default of 1 will be filled in to the start_up_limit or shut_down_limit. Then this check will error. 

## Checklist before merging
- [x] improved [type stability](https://modernjuliaworkflows.org/optimizing/#type_stability) or, when outside the scope of the pull request, at least indicated where and how the code is not properly typed
- [x] Unit tests pass

### DIFF
--- a/src/data_structure/check_data_structure.jl
+++ b/src/data_structure/check_data_structure.jl
@@ -83,7 +83,7 @@ Check if at least one `node` is defined.
 """
 function check_node_object()
     _check(
-        !isempty(node()), "`node` object not found - you need at least one `node` to run a SpineOpt Operations Model",
+        !isempty(node()), "`node` object not found - you need at least one `node` to run a SpineOpt Operations Model"
     )
 end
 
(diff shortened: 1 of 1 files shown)


DillonJ:
@gnawin this causes SpineOpt to fail in our models when neither start_up_limit nor ramp_limt are specified

I wonder why tests were not added?

PR #1312, opened 2026-05-26
Implement #1301

### LINKED ISSUE #1301: Consider making it possible to use ramping capabilities without using online variables
Ramping constraints now include units_on variables with parameters such as start_up_limits, see e.g., https://github.com/spine-tools/SpineOpt.jl/blob/master/src/constraints/constraint_ramp_up.jl.

This is not always convenient for users who only want to model ramping, without needing to set up other parameters.

In the long-term, we should think of how to simplify or create alternatives for this constraint, where only units_on is not needed.

Tasqu:
Fix `_get_units_on` so that it addresses the case with investments but no online variables. Similarly, implement `_get_units_invested_available` to get the investments only if available. Note that explicit `online_variable_type=:none` declarations are required to access this functionality, as otherwise SpineOpt will deduce that online variables are needed.

Fixes #1301 

## Checklist before merging
- [x] Unit tests pass

### DIFF
--- a/src/variables/variable_units_on.jl
+++ b/src/variables/variable_units_on.jl
@@ -108,12 +108,18 @@ function add_variable_units_on!(m::Model)
     )
 end
 
+"""
+    _get_units_on(m, u, s, t)
+
+Safe get `units_on` variable for the given indices,
+replaced with `existing_units` plus [`_get_units_invested_available`](@ref)
+if `units_on` doesn't exist.
+"""
 function _get_units_on(m, u, s, t)
     get(m.ext[:spineopt].variables[:units_on], (u, s, t)) do
-        existing_units(unit=u, stochastic_scenario=s, t=t, _default=_default_nb_of_units(u))
+        (
+            existing_units(unit=u, stochastic_scenario=s, t=t, _default=_default_nb_of_units(u))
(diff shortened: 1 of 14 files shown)


src/variables/variable_units_on.jl
```diff
@@ -108,12 +108,18 @@ function add_variable_units_on!(m::Model)
...
+
+Safe get `units_on` variable for the given indices,
+replaced with `existing_units` plus [`_get_units_invested_available`](@ref)
+if `units_on` doesn't exist.
+"""
 function _get_units_on(m, u, s, t)
     get(m.ext[:spineopt].variables[:units_on], (u, s, t)) do
-        existing_units(unit=u, stochastic_scenario=s, t=t, _default=_default_nb_of_units(u))
+        (
+            existing_units(unit=u, stochastic_scenario=s, t=t, _default=_default_nb_of_units(u))
```
Tasqu, line 127: Include outages?

Tasqu: Check constraint_units_available or something

Tasqu: Done.

Tasqu:
Some caveats remain, see #1301.

PR #1315, opened 2026-06-10
Fix user Constraint fails when using representative periods

### LINKED ISSUE #1314: User Constraint fails with RPs and multi-year investment setup
Issue found when solving the Pan-European Energy System Planning Model

When modeling emissions, I use unit__node__node and fix_ratio_out_in_unit_flow
Example:
- (CCGT, atmosphere, CH4) where fix_ratio_out_in_unit_flow = 0.2 (CO2 ton/ MWh CH4)

But when modeling emissions for a unit with more than one inflows coming from commodities with CO2 content then I use a user_constraint.
Example:
- - Steel production, inputs: coal and waste

There are two relationships unit__from_node
- - (steel_production, coal)
- - (steel_production, waste)
One relationship unit__to_node
- - (steel__production, atmosphere)

User constraint: emission_steel_production
unit__from_node__user_contraint (steel_production, coal, emission_steel_production) with parameter unit_flow_coefficient equals to 0.35 (ton_CO2 / MWh coal)
unit__from_node__user_contraint (steel_production, waste, emission_steel_production) with parameter unit_flow_coefficient equals to 0.3 (ton_CO2 / MWh coal)
unit__to_node__user_contraint (steel_production, atmosphere, emission_steel_production) with parameter unit_flow_coefficient equals to -1

To create the following constraints
0.35 * coal_usage + 0.3 * waste_usage == emission_to_atmosphere

However, I found that the coal flow and waste flow are different from zero but the flow to atmosphere is zero, so that means the constraint is not working.

The only special features I have in the model are the representative periods and the multi-year investment (has free start to True).

Do you think is there something wrong?

datejada:
This pull request includes the following changes:

**Removal of `include_in_non_representative_periods` parameter:**

* The `include_in_non_representative_periods` parameter is removed from all example JSON files, documentation, and the main template, so it is no longer available as a user-configurable option. 

**Codebase updates for constraint enforcement:**

* The function `constraint_user_constraint_indices` is updated to only enforce user constraints in representative periods in the lowest resolution of the variables involved by filtering user-constraint candidate indices to representative temporal blocks before t_lowest_resolution_path.
* The `include_in_non_representative_periods` parameter and its export are removed from `src/convenience_functions.jl`.
* The `_is_representative` function is updated to default to `false` unless explicitly set, ensuring correct identification of representative periods.

**Testing updates:**

* New tests are added to verify that user constraints are only enforced in representative periods in the lowest resolution, including scenarios with and without investment.

Fixes #1314

## Checklist before merging
- [x] improved [type stability](https://modernjuliaworkflows.org/optimizing/#type_stability) or, when outside the scope of the pull request, at least indicated where and how the code is not properly typed
- [x] Unit tests pass

### DIFF
--- a/docs/src/how_to/retrofit_material/retrofitting_example_1.json
+++ b/docs/src/how_to/retrofit_material/retrofitting_example_1.json
@@ -4662,13 +4662,6 @@
         ],
         [
             "user_constraint",
-            "include_in_non_representative_periods",
-            false,
-            "boolean_value_list",
-            "If using representative periods, this boolean parameter determines whether or not the constraint is enforced in non-representative periods."
-        ],
-        [
-            "user_constraint",
             "is_active",
             true,
             "boolean_value_list",
--- a/examples/capacity_planning.json
+++ b/examples/capacity_planning.json
@@ -4367,13 +4367,6 @@
         ],
         [
             "user_constraint",
-            "include_in_non_representative_periods",
-            false,
-            "boolean_value_list",
-            "If using representative periods, this boolean parameter determines whether or not the constraint is enforced in non-representative periods."
-        ],
-        [
-            "user_constraint",
             "right_hand_side",
             0,
             null,
--- a/examples/multi-year_investment_with_econ_discounting_consecutives.json
+++ b/examples/multi-year_investment_with_econ_discounting_consecutives.json
@@ -4217,13 +4217,6 @@
         ],
         [
             "user_constraint",
-            "include_in_non_representative_periods",
-            false,
-            "boolean_value_list",
-            "If using representative periods, this boolean parameter determines whether or not the constraint is enforced in non-representative periods."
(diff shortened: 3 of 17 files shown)


DillonJ:
Nice one @datejada

Tasqu (approved):
Seems to me like the changes are quite small and contained. Adding unit tests is also always appreciated!

I guess the `include_in_non_representative_periods` parameter is vanishing?

datejada:
> Seems to me like the changes are quite small and contained. Adding unit tests is also always appreciated!
> 
> I guess the `include_in_non_representative_periods` parameter is vanishing?

Thanks for the review. The parameter disappears because it no longer plays a role in the new delta formulation (there are no longer replacement expressions for the non-representative periods). So, it is better to take it out to avoid any confusion.

PR #1319, opened 2026-06-18
#480 improved user constraint control

### LINKED ISSUE #480: Retrieving lowest resolution for user constraints
Currently the resolution for creating the user\_constraint is determined based on the resolution of all *potentially* involved variable. For example, looking at the resolution with regards to the unit, we retrieve the lowest resolution ~as follows (simplified):
```
t_lowest_resolution(
    ind.t in unit_flow_indices if u in unit__{from;to\}node__user_constraint(user_constraint=uc);
    ind.t in units_on_indices if u in unit__user_constraint(user_constraint=uc);
    ind.t in units_invested_available_indices if u in unit__user_constraint(user_constraint=uc); 
)
```
However, as both the `units_on` and the `units_invested` variables are associated with the same `user_constraint` relationship, it will always be the lowest resolution of both these variables - even if only one of them is actually included in the constraint. 

I believe an easy fix for the retrieving of the constraint indices could be the following:
```
t_lowest_resolution(
    ind.t in unit_flow_indices if u in unit__{from;to\}node__user_constraint(user_constraint=uc);
    ind.t in units_on_indices if u in unit__user_constraint(user_constraint=uc) && units_on_coefficient(unit=u, user_constraint=uc) != 0;
    ind.t in units_invested_available_indices if u in unit__user_constraint(user_constraint=uc)&& units_invested_available_coefficient(unit=u, user_constraint=uc) != 0;
)
```
A more elegant solution would be to set the default value of both coefficients to `null`. Then the following could be implemented (with some additional adaptions within the constraint):
```
t_lowest_resolution(
    ind.t in unit_flow_indices if u in unit__{from;to\}node__user_constraint(user_constraint=uc);
    ind.t in units_on_indices if u in indices(units_on_coefficient;user_constraint=uc);
    ind.t in units_invested_available_indices if u in indices(units_invested_available_coefficient;user_constraint=uc);
)
```
Any thoughts @DillonJ @manuelma ?

Tasqu:
Change the default values of `user_coefficient` `coefficient_for_*` parameters to `null`, so that we can deduce when such parameters are given intentionally. This, in turn, permits tighter control over the `user_constraint` indexing.

NOTE! The changed default values can always affect the behaviour of old models which have relied on them! Similarly, a few old unit tests had to be updated, as the logic of the user constraint generation is not ever-so-slightly different.

Fixes #480 

## Checklist before merging
- [x] Unit tests pass


Tasqu:
FYI @DillonJ and @manuelma, in case you guys are interested in taking a look.

I was closing old issues today, ran into #480 which I thought would be an easy fix, and fixed it anyway since I had already gotten started. Overall, I think it's still an improvement, but it will change the logic of some old models that had relied on the default value of zero.

DillonJ:
@Tasqu do I understand correctly that the issue only arose when you created a user_constrained and involved only, for example, a units_on coefficient... it would still be created at the investment resolution if that unit had an investment variable?

If that's the case, I'm suprised we haven't come across this before?

test/constraints/constraint_user_constraint.jl
```diff
@@ -244,6 +244,7 @@
             ]
             relationship_parameter_values = [
                 [relationships[1]..., "coefficient_for_node_state", coefficient_for_node_state],
+                [relationships[1]..., "coefficient_for_storages_invested_available", 0.0] # Required for the coarser original resolution caused by the default zeroes.
```
Tasqu, line 247: @DillonJ See example here: The original unit test now needs a separately defined "zero" coefficient for the investment variable, as the old behaviour used the coarser investment resolution even though no coefficient was given by the user, but the `node__user_constraint` was.

Tasqu: The same was presumably true also for `unit__user_constraint`, namely `units_on` and `units_invested`, as you and @mihlema pointed out. However, there is no unit test for that specific case.

test/objective/objective.jl
```diff
@@ -609,21 +609,24 @@ function test_balance_penalty()
...
+    @testset "user_constraint_slack_penalty_original" begin
         url_in = _test_objective_setup()
         uc_slack_penalty = 0.6
         objects = [["user_constraint", "ucx"]]
         relationships = [["node__user_constraint", ["node_a", "ucx"]]]
         object_parameter_values = [
             [objects[1]..., "user_constraint_slack_penalty", uc_slack_penalty],
         ]
+        relationship_parameter_values = [
+            [relationships[1]..., "coefficient_for_demand", 0] # Now needs some coefficient (even zero) to triggen constraint at all.
```
Tasqu, line 621: Also here: If a slack penalty was defined, the original behaviour was to generate user constraints with only slacks, even though no other variables actually appeared in it. Now we need to specifically define a "zero" coefficient for _some_ variable to define the -stochastic-temporal structure if we want to create an "empty" user constraint with only slacks.

PR #1334, opened 2026-08-27
Fix compatibility issues with graph-based SpineInterface

Tasqu:
This PR makes SpineOpt compatible with graph-based SpineInterface.

Also includes possible speed improvements in `constraints/constraints_common.jl`.

Many unit tests now use `with_connection_open()` to make it clear when the database is kept open. Together with the graph-based SpineInterface and latest changes to `spinedb_api` this fixes the random crashes in the unit tests when they are run with multiple threads enabled.

Re spine-tools/SpineInterface.jl#139

## Checklist before merging
- [ ] Unit tests pass


src/variables/variable_node_state.jl
Tasqu, line 1: I don't really like these `intersect`s, as they likely again do set operations on `Vector`s, which might be computationally cumbersome.

test/runtests.jl
Tasqu, line 1: I believe `merge!` here should be replaced with `mergewith!`? Seems like that has been the trend from Julia side.

Tasqu (commented):
I went through the code and didn't find any obvious mistakes. If we can get the tests to pass, that should hopefully be enough.

I did leave two comments, though: Some `intersect`s in the code might be a future point to look at, since set operations over `Vector`s might not be the most efficient. Also, `merge!` should likely be replaced with `mergewith!`, as that seems to have been the trend with Julia so far.

src/variables/variable_node_state.jl
Tasqu: Long-term todo.

PR #1342, opened 2026-09-16
Precompile

tarskul:
Precompile SpineOpt with PrecompilationTools. Precompilation bridges the gap between the theoretical speed of the program and the practical speed; i.e. the first and second run of SpineOpt in the julia REPL. Precompilation is never as good as actually writing performant type stable code but it is a good solution while we are growing towards the more performant code. Keep in mind that when we are writing the more performant code, for a long while we will not see the benefits compared to the precompilation but that does not mean that we don't need to pursue that.

Precompilation also does not cover every use case whereas performant code does. However, this can be mitigated by improving the case that is used for the precompilation. If we notice that a constraint is missing, we can add it to the precompile example. For now, the precompile example is based on the capacity planning example.


tarskul:
Currently waiting for the SpineInterface version to be bumped.

tarskul:
Apparently I initially only tested for julia 1.12. The latest changes now also ensure that it works up to julia 1.10.

